Issue #7, opened 2025-04-08
ebay-snackbar-dialog: don't override input.open set to false


janenath:
# Bug Report

## Description

Updating `ebay-snackbar-dialog` open input to `false`, `<ebay-snackbar-dialog open=state.open />`, does not close the snackbar component. On inspection it seems [this line](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-snackbar-dialog/component.ts#L38) is overriding `input.open` to use `this.state.open` if `input.open` is falsy, preventing the snackbar from closing.

## Screenshots

<img width="941" alt="Image" src="https://github.com/user-attachments/assets/8d32e23e-0a34-4c36-8340-c219e8e33e9b" />

PR #10, opened 2025-04-24
fix(snackbar): open is honored from input and accesskey is spread correctly

### LINKED ISSUE #3: ebay-snackbar-dialog: action accesskey property not being set on <button/>
# Bug Report

## Description

* `ebay-snackbar-dialog` let's you pass an `@action` custom attribute which will add a `ebay-fake-link`
* the `accessKey` param is not passed through to the `ebay-fake-link` which means the keyboard shortcut control + option + accessKey (Mac) doesn't work

## Workaround

* tab navigate to the snackbar until the button is focused and trigger with space or enter (potentially hard to keyboard navigate to snackbar within the 6s snackbar duration depending on DOM structure)

## Fix

Add
```js
accesskey=action.accessKey
```
here: https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-snackbar-dialog/index.marko#L25-L28

## Screenshots

Image

agliga:
Fixes #3 #7

## Description
* Added a fix to allow open to swap based on input
* Added a fix to pass in actions to fake-button

## Checklist

### DIFF
--- a/.changeset/public-hands-relate.md
+++ b/.changeset/public-hands-relate.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+fix: honor snackbar dialog input.open if type boolean
--- a/packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
+++ b/packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
@@ -35,7 +35,12 @@ class SnackbarDialog extends Marko.Component<Input, State> {
     }
 
     onInput(input: Input) {
-        this.state = { open: input.open || this.state.open || false };
+        this.state = {
+            open:
+                typeof input.open === "boolean"
+                    ? input.open
+                    : this.state.open || false,
+        };
     }
 
     onMount() {
--- a/packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
+++ b/packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
@@ -5,6 +5,8 @@ import type { Input as BaseInput } from "../components/ebay-dialog-base/componen
 interface SnackbarDialogInput extends Omit<BaseInput, `on${string}`> {
     layout?: "row" | "column";
     action?: BaseInput["action"] & {
+        accesskey?: string;
+         /** @deprecated use `accesskey` instead */
         accessKey?: string;
     };
     "on-action"?: () => void;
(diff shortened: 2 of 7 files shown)


.changeset/public-hands-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
```
LuLaValva, line 2: Double changeset? You should probably remove one of these.

packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
```diff
@@ -5,7 +5,7 @@ import type { Input as BaseInput } from "../components/ebay-dialog-base/componen
 interface SnackbarDialogInput extends Omit<BaseInput, `on${string}`> {
     layout?: "row" | "column";
     action?: BaseInput["action"] & {
-        accessKey?: string;
+        accesskey?: string;
```
LuLaValva, line 8: Will this be a breaking change for TS users? Maybe we should support both, and derive `accesskey` from `accessKey`

packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
```diff
@@ -35,7 +35,12 @@ class SnackbarDialog extends Marko.Component<Input, State> {
     }
 
     onInput(input: Input) {
-        this.state = { open: input.open || this.state.open || false };
+        this.state = {
+            open:
+                typeof input.open === "boolean"
+                    ? input.open
+                    : this.state.open || false,
+        };
```
LuLaValva, line 43: This is what nullish coalescing is for! You can use
```suggestion
        this.state = { open: input.open ?? this.state.open || false };
```

.changeset/public-hands-relate.md
agliga: There are two changes in this, two issues

packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
agliga: I believe since this was not working at all, we should change it to the right attribute.

packages/ebayui-core/src/components/ebay-snackbar-dialog/component.ts
```diff
@@ -5,7 +5,8 @@ import type { Input as BaseInput } from "../components/ebay-dialog-base/componen
 interface SnackbarDialogInput extends Omit<BaseInput, `on${string}`> {
     layout?: "row" | "column";
     action?: BaseInput["action"] & {
-        accessKey?: string;
+        accesskey?: string;
+        accessKey?: string; // deprecated
```
LuLaValva, line 9: Can you use JSDoc here, so people see deprecations from their editor?
```suggestion
        /** @deprecated use `accesskey` instead */
        accessKey?: string;
```

LuLaValva (approved):

PR #122, opened 2025-05-13
feat(table): added frozen header option

agliga:
Fixes #

## Description
* Added `frozen-header` option to table in ebayui core
* Added example and test

## Screenshots
Screenshot 2025-05-13 at 8 55 26 AM

## Checklist

### DIFF
--- a/packages/ebayui-core/src/components/ebay-table/examples/frozen-header.marko
+++ b/packages/ebayui-core/src/components/ebay-table/examples/frozen-header.marko
@@ -1,0 +1,46 @@
+import data from "./data.json";
+
+<ebay-table style={height: "300px"} frozen-header ...input>
+    <@header row-header>
+        Seller
+    </@header>
+    <@header>Item</@header>
+    <@header>Status</@header>
+    <@header column-type="numeric">
+        List Price
+    </@header>
+    <@header column-type="numeric">
+        Quantity Available
+    </@header>
+    <@header>Orders</@header>
+    <@header column-type="numeric">
+        Watchers
+    </@header>
+    <@header column-type="numeric">
+        Protection
+    </@header>
+    <@header>Shipping</@header>
+    <@header>Delivery</@header>
+    <for|r| of=data>
+        <@row>
+            <@cell>${r.seller}</@cell>
+            <@cell>${r.item.title}</@cell>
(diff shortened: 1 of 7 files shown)


saiponnada (commented):
One issue I noticed is regarding the focus being obscured as part of the frozen header. I have noticed this issue on ebayui-core is intermittent. I would recommend instead of browser controlling the scroll, we have our own logic to make sure focus is not obscured.

packages/ebayui-core/src/components/ebay-table/examples/frozen-header.marko
```diff
@@ -0,0 +1,46 @@
+import data from "./data.json";
+
+<ebay-table style={height: "300px"} frozen-header ...input>
```
ArtBlue, line 3: The inline style in Skin was only to allow the frozen header to be demonstrated, so we should remove it from here.

While we're at it, can you remove it from the example code in Skin?

packages/ebayui-core/src/components/ebay-table/table.stories.ts
```diff
@@ -192,3 +198,14 @@ export const ColumnSortingClientSide = buildExtensionTemplate(
...
     sortClientSideCode,
 );
+
+export const FrozenHeader = buildExtensionTemplate(
+    frozenHeaderTemplate,
+    frozenHeaderCode,
+    {
+        frozenHeader: true,
+        style: {
+            height: "300px",
```
ArtBlue, line 208: Don't need the height here either.

ArtBlue (changes requested):
Some minor things...

packages/ebayui-core/src/components/ebay-table/examples/frozen-header.marko
ArtBlue: Ah...it looks like I missed the max-height in Skin. 

<kbd>image</kbd>

ArtBlue: I made the Skin fixes, so these can be removed.

agliga:
Done, I went ahead and removed the frozen header example. No need since its just a toggle

agliga:
React changes LGTM

PR #138, opened 2025-05-20
feat: Add EbayItemTile and EbayItemTileGroup Component - wip

bruno2kd:
Add EbayItemTile and EbayItemTileGroup Component

- Waiting for PR https://github.com/eBay/evo-web/pull/140 to be merged

## Description
Add EbayItemTile Component

## Screenshots
Screenshot 2025-05-20 at 17 42 41
Screenshot 2025-05-20 at 17 42 32

## Checklist


packages/ebayui-core-react/src/ebay-file-preview-card/file-preview-action.tsx
```diff
@@ -81,14 +91,27 @@ const EbayFilePreviewAction: FC<EbayFilePreviewActionProps> = ({
...
-        />
-    )
+
+    if (isValidElement(children) && children?.props['aria-label']) {
+        const child = children as ReactElement<ComponentProps<'button'>>
+        return cloneElement(child, {
+            onClick: onAction,
+            className: cx('file-preview-card__action', child?.props?.className)
+        })
+    }
```
HenriqueLimas, line 101: This logic doesn't looks correct and over using react helpers. We can instead return this as the default return bellow and passing the props to the EbayIconButton instead and passing the `icon` and the rest of the params that is being passed to this components.

```
if (deleteText) {
  return ...
}

return (
  <EbayIconButton className={cx(className, 'file-preview-card__action')} {...rest} />
)
```

The react EbayIconButton is a little bit different on how the icon is being passed. On marko it is a children, in react we pass a `icon` string,

packages/ebayui-core-react/src/ebay-file-preview-card/file-preview-action.tsx
```diff
@@ -7,14 +14,15 @@ import {
     FilePreviewCardMenuActionHandler
 } from './types'
 
-export type EbayFilePreviewActionProps = {
+export type EbayFilePreviewActionProps = ComponentProps<'button'> & {
```
HenriqueLimas, line 17: It should be IconButtonProps instead

packages/ebayui-core-react/src/ebay-item-tile/item-tile.tsx
```diff
@@ -0,0 +1,68 @@
...
+    layout,
+    className,
+    onAction,
+    children,
+    ...rest
+}) => {
+    const supertitle = findComponent(children, EbayItemTileSectionSuperTitle)
+    const title = findComponent(children, EbayItemTileSectionTitle)
+    const subtitle = findComponent(children, EbayItemTileSectionSubTitle)
+    const action = findComponent(children, EbayIconButton)
```
HenriqueLimas, line 35: We should have a `EbayItemTileAction` instead

packages/ebayui-core-react/src/ebay-file-preview-card/file-preview-card.tsx
```diff
@@ -39,12 +43,16 @@ const EbayFileInput: FC<EbayFilePreviewCardProps> = ({
...
+    href,
     onCancel,
     onDelete,
     onMenuAction,
     onSeeMore,
+    onAction,
+    children,
     ...rest
 }) => {
+    const action = findComponent(children, EbayIconButton)
```
HenriqueLimas, line 55: We should use `EbayFilePreviewAction` instead of EbayIconButton

packages/ebayui-core-react/src/ebay-item-tile/item-tile.tsx
```diff
@@ -0,0 +1,68 @@
...
+            {...rest}
+        >
+            {file && (
+                <div className="item-tile__header">
+                    <EbayFilePreviewCard
+                        file={file}
+                        href={href}
+                        onAction={onAction}
+                    >
+                        {action}
```
HenriqueLimas, line 52: Considering the comments above, this will be:
```jsx
<EbayFilePreviewCard ...>
  <EbayFilePreviewCardAction {...action.props} />
</EbayFilePreviewCard>
```

HenriqueLimas (changes requested):
Thanks @bruno2kd for the PR. can you separate the PR so we can have release notes separately for each specific component and it is easier to review? I added some initial changes for now.

packages/ebayui-core-react/src/ebay-file-preview-card/file-preview-action.tsx
bruno2kd: I see, so we should have a Icon type prop which will go to EbayIconButton in the component.

About the code you used as example, In Marko's implementation if action and deleteText is passed the component shows the action and also if nothing is passed, there is no action and component only display the image.

bruno2kd: This is the action description in the [storybook](https://opensource.ebay.com/evo-web/ebayui-core/?path=/story/layout-ebay-item-tile--default), that is why I checked for the `aria-label`:

@action | The icon to be used for the action button in the header. An aria-label is also required for accessibility. If not provided, the action button will not be rendered.

PR #144, opened 2025-05-22
fix: on close bug

ericbiewener:
Fixes # N/A

- [x] This PR does not contain CSS changes

## Description

Under certain conditions, clicking a button inside a dialog causes the dialog to incorrectly close. A key condition that must be met is that the clicked element disappears from the DOM after being clicked. That causes [this condition](https://github.com/eBay/evo-web/blob/f52a6dded13308d7bb569cf5049b19cf85e34b53/packages/ebayui-core-react/src/ebay-dialog-base/components/dialogBase.tsx#L101) to pass because `!drawerBaseEl.current.contains(e.target)` evaluates to true since `e.target` has been removed from the DOM.

This PR adds an additional check to that conditional to validate that `e.target` is still _somewhere_ in the document. If it's not, then the dialog does not close. If it _is_ still in the document, then that indicates that the user indeed clicked outside the dialog, and thus it should close. 

For some reason I have been unable to repro this bug with a smaller use case, so I'm not sure what other conditions contribute to it in the first place. But regardless, this fix seems appropriate.

## Checklist

- [ ] I verify the build is in a non-broken state

### DIFF
--- a/packages/ebayui-core-react/src/ebay-dialog-base/components/dialogBase.tsx
+++ b/packages/ebayui-core-react/src/ebay-dialog-base/components/dialogBase.tsx
@@ -98,7 +98,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
     useEffect(() => {
         let timeout: number
         const handleBackgroundClick = (e: DocumentEventMap['click']) => {
-            if (drawerBaseEl.current && !drawerBaseEl.current.contains(e.target)) {
+            if (drawerBaseEl.current && !drawerBaseEl.current.contains(e.target) && document.contains(e.target as Node)) {
                 onBackgroundClick(e as unknown as DialogCloseEvent)
             }
         }
--- a/packages/ebayui-core-react/src/ebay-lightbox-dialog/__tests__/index.stories.tsx
+++ b/packages/ebayui-core-react/src/ebay-lightbox-dialog/__tests__/index.stories.tsx
@@ -1,19 +1,14 @@
-import React, { useState } from 'react'
-import { Meta, StoryFn } from '@storybook/react'
-import { action } from '@storybook/addon-actions'
-import {
-    EbayDialogFooter,
-    EbayDialogHeader,
-    EbayDialogPreviousButton
-} from '../../ebay-dialog-base'
-import { EbayButton } from '../../ebay-button'
-import { EbayCheckbox } from '../../ebay-checkbox'
-import { EbayLabel } from '../../ebay-field'
-import { EbayLightboxDialog } from '../index'
-
+import React, { useState } from "react";
+import { Meta, StoryFn } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import { EbayDialogFooter, EbayDialogHeader, EbayDialogPreviousButton } from "../../ebay-dialog-base";
+import { EbayButton } from "../../ebay-button";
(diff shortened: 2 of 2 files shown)


packages/ebayui-core-react/src/ebay-lightbox-dialog/__tests__/index.stories.tsx
ericbiewener, line 1: Lots of formatter prettier changes in this file. Not going to fix it since it doesn't successfully repro the failure case this PR solves. The `DebugOnClose` story at the end of this file was my attempt to fix it.

HenriqueLimas: can you undo the Prettier changes for this file? I think your IDE is autoformatting using prettier but we don't use it in our project (yet)

packages/ebayui-core-react/src/ebay-dialog-base/components/dialogBase.tsx
```diff
@@ -98,7 +98,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
     useEffect(() => {
         let timeout: number
         const handleBackgroundClick = (e: DocumentEventMap['click']) => {
-            if (drawerBaseEl.current && !drawerBaseEl.current.contains(e.target)) {
+            if (drawerBaseEl.current && !drawerBaseEl.current.contains(e.target) && document.contains(e.target as Node)) {
```
HenriqueLimas, line 101: Can we add a comment for why we are adding document.contains. Also, I think you can check by `e.target.parentNode`, if it is `null` or empty it means its not on the dom and might be faster then `contains`

agliga:
Im actually wondering if this is actually an issue and even if it seems appropriate, I am reluctant to accept it. 
What is the usecase where a button is triggering a close on the dialog, when it isn't even in the DOM? I mean why not remove the button from the dom AFTER all the events of the dialog occur.
Before we even look at this PR, we need to reproduce this in storybook and have a clear understanding of what is happening and we need to accept it. Sure a small fix like this can be appropriate, but it can lead to other bugs or other side effects. 
Lets make sure first.

packages/ebayui-core-react/src/ebay-lightbox-dialog/__tests__/index.stories.tsx
```diff
@@ -357,7 +352,36 @@ export const Expressive: StoryFn<typeof EbayLightboxDialog> = (args) => {
...
+            <button className="btn btn--secondary" onClick={() => setOpen(!open)}>
+                Open Dialog
+            </button>
+            <EbayLightboxDialog
+                {...args}
+                bannerImgSrc="http://ir.ebaystatic.com/cr/v/c1/skin/image-treatment/mountain.jpeg"
+                bannerImgPosition="top"
+                open={open}
+                onOpen={() => action("onOpen")()}
+                onClose={() => {
```
HenriqueLimas, line 373: This onClose function is causing not being able to simulate the issue. We need to wrap it in a `useCallback()` so it will not change the reference on every render and the `useEffect()` will not trigger the removeEventListener

HenriqueLimas:
@agliga The issue is caused when the element that is being clicked it is also being removed from the DOM. It wasn't simulating because of a small detail on the test case, but it is a valid issue.

PR #146, opened 2025-05-23
chore: added visual regression testing for coreui

agliga:

## Description
* Added visual regression testing for coreui 
* Created some shortcut methods to allow easy testing (such as diff snapshot testing)

This is a sample of what the visual regression test would be using visual-html. I only changed ccd (which had no browser tests, all it does is take a snapshot of all the styles) and accordion (which had some browser checks, but used visual html and diffing tools to verify the changes).

## Notes
* There is a bug with visual HTML on button currently, working to figure out what the issue is.

## Checklist

### DIFF
--- a/packages/ebayui-core/src/components/ebay-accordion/test/__snapshots__/test.browser.js.snap
+++ b/packages/ebayui-core/src/components/ebay-accordion/test/__snapshots__/test.browser.js.snap
@@ -22,6 +22,8 @@ exports[`accordion > given the accordion in the default state > then it matches 
       transition-duration: 0.1s
     ">
       <details style="
+        overflow-x: hidden;
+        overflow-y: hidden;
         transition-delay: 0.1s;
         transition-duration: 0.1s
       ">
@@ -69,6 +71,8 @@ exports[`accordion > given the accordion in the default state > then it matches 
             <svg
               focusable="false"
               style="
+                animation: chevron-bounce-close var(--motion-duration-medium-1)
+                        var(--motion-easing-standard);
                 display: inline-block;
                 fill: var(
                         --details-secondary-foreground-color,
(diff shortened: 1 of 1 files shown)


ArtBlue:
This looks cool! Since there's no UI tooling for this, would it just trigger a failure from the script to let us know of a potential diff issue?

agliga:
> This looks cool! Since there's no UI tooling for this, would it just trigger a failure from the script to let us know of a potential diff issue?

So it will show the file diff, namely the styles that are missing from each.

agliga:
For example, I went in skin and removed `width: 100%` from accordion
After running the build I get this error:
```
FAIL   browser (chromium)  src/components/ebay-accordion/test/test.browser.js > accordion > given the accordion in the default state > then it matches the snapshot
Error: Snapshot `accordion > given the accordion in the default state > then it matches the snapshot 1` mismatched

- Expected
+ Received

@@ -5,12 +5,11 @@
    <ul style="
      margin-bottom: 0px;
      margin-top: 0px;
      padding-left: 0px;
      transition-delay: 0.1s;
-     transition-duration: 0.1s;
-     width: 100%
+     transition-duration: 0.1s
    ">
      <style scoped>
        ::marker {font-size: 0px}
      </style>
      <li style="

 ❯ src/components/ebay-accordion/test/test.browser.js:41:26
     39|             const html = visualHTML(component.container);
     40|
     41|             expect(html).toMatchSnapshot();
       |                          ^
     42|         });
     43|
```
So while we don't see the difference, we see the actual styles being applied. Good news is this is only one snapshot, the others pass since they only do a diff between their changes (IE, get a snapshot, click and element, get another snapshot, diff the first and last)
I can add a update snapshot command (which is pretty straightforward anyways) whenever skin updates and causes coreui snapshots to fail. But we can discuss if we want to take a full snapshot or just the diff when actions happens. (I have the base case takes a full snapshot, while any interactions only take diffs)

LuLaValva (dismissed):

LuLaValva (approved):

PR #168, opened 2025-06-06
feat(chips): added new chips

agliga:
Fixes #

## Description
This is a draft of the chips implementation

This basically takes current chips and adds interactive chips to them.

This is the new set of chips:
* static chip - black background with no 8px border-radius, existing
* desctructive chip - same as static chip but has a close button, existing
* interactive chip - toggles between black background and white, 8px border radius
* filter chip - use `chip--fitler` modifier. Adds a 16px border radius, but unchanged from other chips otherwise (other than having a leading icon)
* discovery-chip - use `chip--discovery` (I dont remember the new name, will rename this). Has 16px and larder height. Unchanged from chips (other than leading media container)


agliga:
> Nice overall! This does feel like the right approach overall - everything under chip, and we possibly sunset the redundant components.

Thanks!
Yes this is a very rough implementation without all the examples set. Part of it is that I wanted to get early feedback on this approach. I didn't want to go and add all the changes and then have to completely redo it and have more throwaway code :) 

If we are fine with it, I will go ahead and clean it up and fix all the examples and such.

packages/skin/src/routes/_index/component/chip/+page.marko
```diff
@@ -26,10 +26,366 @@
...
+            >
+                <img
+                    class="chip__media"
+                    src="https://ir.ebaystatic.com/cr/v/c01/skin/docs/dog_profile_optimized.jpg"
+                    alt=""
+                >
+                <span class="chip__text">
+                    Football
+                    <span class="clipped">
+                        - selected
```
ianmcburnie, line 346: For this text in all cases, lets go with " - filter applied". This is what we use on the SRP left rail filters. This also avoids any overlap with ARIA terminology (e.g. "selected")

packages/skin/src/routes/_index/component/chip/+page.marko
```diff
@@ -26,10 +26,366 @@
         </span>
     </highlight-code>
 
-    <h3 id="interactive-chip">
-        Interactive Chip
+    <h3 id="toggle-chip">
+        Toggle Chip
     </h3>
-    <p>An interactive chip has a delete button inside for chip removal.</p>
+    <p>An toggle chip toggles between two states, selected and unselected.</p>
```
ianmcburnie, line 32: selected and unselected are for th `aria-selected` propery. Let's make sure all references are updated and correct for aria-pressed.

packages/skin/src/routes/_index/component/chip/+page.component-browser.js
```diff
@@ -0,0 +1,20 @@
+function toggleAttribute(chip, attribute) {
```
ArtBlue, line 1: Can't we just use the native js toggle? https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute

packages/skin/src/routes/_index/component/chip/+page.marko
```diff
@@ -26,10 +26,287 @@
...
+            <span class="chip__text">
+                Football
+                <span class="clipped">
+                    - selected
+                </span>
+            </span>
+        </a>
+    </highlight-code>
+
+    <h3 id="destructive-chip">
```
ArtBlue, line 304: Since destructive was original to chip and seems to be simpler than the others, can we move this one to the top, right after the static chip?

packages/skin/src/routes/_index/component/chip/+page.marko
```diff
@@ -67,16 +339,83 @@
...
             >
+                <svg class="icon icon--12" width="13" height="12">
+                    <use href="icons.svg#icon-close-12"/>
+                </svg>
+            </button>
+        </span>
+    </highlight-code>
+
+    <h3 id="interactive-chip">
+        Menu button Chip
```
ArtBlue, line 350: Should we do title case?

packages/skin/dist/chip/chip.css
```diff
@@ -21,12 +89,16 @@
 }
 
 button.chip__button {
```
ArtBlue, line 91: This probably needed cascading from the component originally. Should we change it? 

`.chip .chip__button`

agliga: I believe this won't work because we prefix all button styles with `button.btn` (this is because it can be an anchor or a button)
I will prefix it with `.chip` so that the order the CSS won't cause issues.

packages/skin/src/routes/_index/component/chip/+page.component-browser.js
agliga: Will change to it, but will need to keep check still (so that if aria-expanded is not set, we dont toggle it)

agliga: Actually this doesn't work. It only toggles between `aria-pressed` set or unset. We want `aria-expanded="true"` or `aria-expanded="false"` (string values)

packages/skin/src/routes/_index/component/selection-chip/+page.marko
```diff
@@ -0,0 +1,43 @@
+<div id="chip">
+    <section-header metadata=metadata/>
+
+    <h3 id="selection-chip">
+        Selection Chip
+    </h3>
+    <p>An selection chip toggles between two states, selected and unselected.</p>
+    <p>A selection chip is used in a form as a checkbox. This is generally used to select predetermined values</p>
```
ianmcburnie, line 8: This doesn't seem right. It's not really a checkbox. May be best to just delete this line. I can actually sync up these parts with the MIND pattern docs if you like.

agliga: Sure. I haven't updated the text for this yet but if you want to I'm fine with that

ianmcburnie: Done

ArtBlue (approved):
LGTM!

ianmcburnie:
This image and text feels out of place (sorry doggy!) Maybe grab an example from the figma. Or just change the text to "Pet Supplies" maybe.

Screenshot 2025-07-09 at 2 57 27 PM

agliga:
> This icon spins when I load the page. Is that intentional?
> 
> Screenshot 2025-07-09 at 3 03 02 PM

Its the animation for this when it loads. I think there's no real way around it unless we do javascript animations.

agliga:
> Did you check the colour contrast of this kb focus outline? Might need switching to white here.
> 
> Screenshot 2025-07-09 at 2 53 52 PM

I believe this is just the "default" focus. While it could be a color contrast issue, there aren't any specs for this. 

> Did you check the colour contrast of this kb focus outline? Might need switching to white here.
> 
> Screenshot 2025-07-09 at 2 53 52 PM

So in the spec they have the outline on the whole thing. We can swap to a white color but there's no spec for that currently. This probably should be something we should look at after we change all the focus states since this is just using the default.

ArtBlue (changes requested):
It's looking really good. A couple of minor things I fount.

1. The menu filter does not include the listbox. I think it should. The interactive component will need it as well. Maybe it does exist in the `css` though it doesn't look like it. In the docs, we should open the menu when users click to let them see the visuals inside. The selection function doesn't necessarily need to be there, but we should have all the states accounted for visually. The storybook is also missing the expanded listbox.

<kbd>Image</kbd>

2. The selection chip looks broken in storybook. Maybe the markup wasn't updated with the docs? 

<kbd>Image</kbd>

agliga:
> It's looking really good. A couple of minor things I fount.
> 
> 1. The menu filter does not include the listbox. I think it should. The interactive component will need it as well. Maybe it does exist in the `css` though it doesn't look like it. In the docs, we should open the menu when users click to let them see the visuals inside. The selection function doesn't necessarily need to be there, but we should have all the states accounted for visually. The storybook is also missing the expanded listbox.
> 
> Image
> 

So as I mentioned before, the menu part of this will be built in a second pass.
This only has the button portion. For the next pass, we will integrate it with menu button, where we will have the dropdown attached to it. 

> 2. The selection chip looks broken in storybook. Maybe the markup wasn't updated with the docs?
> 
> Image

Yeah, this is an issue with animations. Will update.

PR #181, opened 2025-06-16
chore(ai): added the initial shared instructions for copilot

ArtBlue:
## Shared Copilot Instructions

This is for adding a shared set of Copilot instructions for use on local machines.

### How to Use

1. Click on "Add Context"
2. From the selection, select "Instructions..."
3. Load the prompt instruction file(s) from `.github/instructions`

Your Copilot interactions will then take the instructions as context.

### Updates

Feel free to suggest changes. These were just the initial set of instructions that came to mind for the `scss` and `html`.

### DIFF
--- a/.github/instructions/skin-css-prompt.instructions.md
+++ b/.github/instructions/skin-css-prompt.instructions.md
@@ -1,0 +1,18 @@
+---
+applyTo: "**/*.scss"
+---
+
+Coding standards, domain knowledge, and preferences that AI should follow.
+
+Always use BEM (Block Element Modifier) naming conventions for all CSS/SCSS selectors.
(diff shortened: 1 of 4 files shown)


ianmcburnie:
Should we have a single source of truth for all IDEs? e.g. Cline, Cursor, etc

ianmcburnie:
Should we point it to our skin style guide?

ArtBlue:
> Should we also add a section for .js or .ts files?

@agliga , absolutely. This is just the html/css/scss peace. We can iterate in the future and include any additional instructions specific to various other types of files.

agliga (approved):
LGTM, we can always iterate if there are issues.

PR #185, opened 2025-06-23
feat(avatar): added new evo-avatar component

agliga:
Added avatar component in marko 6


packages/evo-marko/src/tags/evo-avatar/component.ts
```diff
@@ -0,0 +1,44 @@
+import type { WithNormalizedProps } from "../../global";
```
LuLaValva, line 1: This whole component file should be removed, right? I think all of the logic was moved into `index.marko`

PR #188, opened 2025-06-24
Adding value into pagination item

bsalescardoso:

## Description
This change allows you to set the `value` of each EbayPaginationItem, then the `onSelect` even will have the proper identification of the page, this will allow have a different value than the representation of the page number (isn't a problem right now because we don't have the website on countries that numeral representation is different than latin)

Also this is fixing the README about `value` vs `index` and including the newish property documentation.

### DIFF
--- a/.changeset/loud-dogs-dig.md
+++ b/.changeset/loud-dogs-dig.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ui-core-react": minor
+---
+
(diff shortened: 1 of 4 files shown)


bsalescardoso:
There is some tests failing on `ebayui-core/src/components/ebay-date-textbox/test/test.server.js` which sounds to be localisation + timezone related.

Screenshot 2025-06-24 at 14 10 24

PR #190, opened 2025-06-24
feat(icons): converted icons to marko 6

agliga:


packages/evo-marko/src/tags/evo-icon/index.marko
```diff
@@ -0,0 +1,105 @@
...
+<id/titleId/>
+<const/prefixSvg=(_type === "icon" ? "icon-" : "")/>
+<const/hasA11yText=(a11yText || input["aria-label"])/>
+<const/hasTitle = (a11yText && !input["aria-label"])/>
+
+<svg
+    ...htmlInput
+    class=[getIconClass(input._type ?? "", input._name, input._size)]
+    focusable="false"
+    role=(hasA11yText && "img")
```
LuLaValva, line 93: TypeScript doesn't like this because `hasA11yText` can be `""`, can you cast it to a boolean on line 86 with `!!`?

packages/evo-marko/src/tags/evo-icon/index.marko
```diff
@@ -0,0 +1,105 @@
...
+
+<id/titleId/>
+<const/prefixSvg=(_type === "icon" ? "icon-" : "")/>
+<const/hasA11yText=(a11yText || input["aria-label"])/>
+<const/hasTitle = (a11yText && !input["aria-label"])/>
+
+<svg
+    ...htmlInput
+    class=[getIconClass(input._type ?? "", input._name, input._size)]
+    focusable="false"
```
LuLaValva, line 92: This seems like it's legacy from IE, we can get rid of it now that IE is deprecated:

https://html-validate.org/rules/svg-focusable.html

packages/evo-marko/src/tags/evo-icon/index.marko
```diff
@@ -0,0 +1,104 @@
...
+        } else {
+            return symbol();
+        }
+    }
+
+    return "";
+}
+
+<id/titleId/>
+<const/prefixSvg=(_type === "icon" ? "icon-" : "")/>
```
LuLaValva, line 85: Looks like this is unused

LuLaValva (commented):
Looks good! Just one unused attribute to remove and then it should be good to merge. Unless we want to move forward with changing `<evo-[name]-[size]-icon>` to `<evo-icon-[name]-[size]>` (which I am a strong advocate for)

agliga:
> Looks good! Just one unused attribute to remove and then it should be good to merge. Unless we want to move forward with changing `<evo-[name]-[size]-icon>` to `<evo-icon-[name]-[size]>` (which I am a strong advocate for)

We can easily swap to the new syntax before we release. So lets leave it as is for now.

PR #191, opened 2025-06-24
feat(3d-viewer): added new component in marko6

agliga:
Will need to fix icons here but will do once icon PR is merged

### DIFF
--- a/packages/evo-marko/src/tags/evo-3d-viewer/index.marko
+++ b/packages/evo-marko/src/tags/evo-3d-viewer/index.marko
@@ -49,7 +49,7 @@ static type LoadState = "ready" | "error" | "loading";
     </if>
     <else if=loadState === "error">
         <div class="three-d-player__overlay">
-            // <ebay-attention-24-icon/>
+            // <evo-attention-24-icon/>
             <div class="three-d-player__overlay-text">
                 ${errorText}
             </div>
--- a/packages/evo-marko/src/tags/evo-3d-viewer/index.marko
+++ b/packages/evo-marko/src/tags/evo-3d-viewer/index.marko
@@ -1,0 +1,81 @@
+import type { AttrClass, AttrString } from "marko/tags-html";
+import type { ModelViewerElement } from "@google/model-viewer";
+import { load as modelViewerLoad } from "@internal/model-viewer";
+export interface Input extends Partial<ModelViewerElement> {
+    class?: AttrClass;
+    errorText?: AttrString;
+    a11yLoadText?: AttrString;
+    onLoadError?: (err: CustomEvent) => void;
+}
+<const/{
+    class: inputClass,
+    src,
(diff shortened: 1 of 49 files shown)


packages/evo-marko/src/tags/evo-progress-spinner/progress-spinner.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+        ariaLabel: {
+            control: { type: "text" },
+            description: "Description for accessibility",
+        },
+    },
+};
+
+export const Default = Template.bind({});
+Default.args = {
+    ariaLabel: "loading",
```
LuLaValva, line 37: This should be `"aria-label"`

packages/evo-marko/src/tags/evo-progress-spinner/examples/large.marko
```diff
@@ -0,0 +1 @@
+<evo-progress-spinner size="large" aria-label="Busy"/>
```
LuLaValva, line 1: Should we add `...input` here so it's adjustable in Storybook?

In general, should we be adding `...input` to these examples? I fear that devs might copy-paste them when spreading input is not actually a good idea, but maybe that's not justified

packages/evo-marko/src/tags/evo-3d-viewer/index.marko
```diff
@@ -0,0 +1,81 @@
...
+    class: inputClass,
+    src,
+    errorText,
+    onLoadError,
+    a11yLoadText,
+    ...rest
+}) => rest)(input)/>
+
+<let/ready=false/>
+<let/error=false/>
```
LuLaValva, line 29: Can we ever be ready and error at the same time? If not, this should probably be `<let/state="pending" as "ready" | "error" | "pending">`

packages/evo-marko/src/tags/evo-3d-viewer/index.marko
```diff
@@ -0,0 +1,81 @@
+import type { AttrClass, AttrString } from "marko/tags-html";
+import type { ModelViewerElement } from "@google/model-viewer";
+import { load as modelViewerLoad } from "@internal/model-viewer";
+export interface Input extends Partial<ModelViewerElement> {
+    class?: AttrClass;
+    errorText?: AttrString;
+    a11yLoadText?: AttrString;
```
LuLaValva, line 7: Now that we don't do any automatic case conversion, it's safe to make attributes mandatory. Should we be enforcing `a11yLoadText` and `errorText` with TypeScript?

packages/evo-marko/src/tags/evo-progress-spinner/examples/large.marko
agliga: Well we do that so it works in storybook....

PR #193, opened 2025-06-24
feat(react): add EbayFlag components for SVGs

### LINKED ISSUE #95: Add EbayFlag component
https://opensource.ebay.com/ebayui-core/?path=/docs/graphics-icons-ebay-flag--documentation

HenriqueLimas:
Fixes #95 

- [x] This PR does not contain CSS changes

## Description

- Add new `EbayFlag` component that follows our `EbayIcon` component.
- Update icon scripts to consider flags symbols and types generation
- Add `flags` property on `EbaySvg` component to render flags symbols when enabled

## Notes

`EbayFlag` is different then the marko `<ebay-flag>` component, so we follow the same approach that we use on `EbayIcon` where we don't have a different component for each icon. This is to reduce the bundle size in React, where on Marko that is handle and minimized by the compiler.

## Screenshots

Screenshot 2025-06-24 at 1 46 10 PM

## Checklist

- [ ] I verify the build is in a non-broken state


packages/ebayui-core-react/.storybook/preview.jsx
```diff
@@ -10,7 +10,7 @@ export default {
     decorators: [
         (Story) => (
             <StrictMode>
-                <EbaySvg />
+                <EbaySvg flags />
```
agliga, line 13: Maybe it might be better to have this as a separate component, such as `EbayFlagsSvg`. It seems small enough.

agliga (commented):
One more comment, we actually have two flag types, one is the SVG which is used mainly for high rez flags. The other is a sprite and thats recommended for things like phone input or when you want to switch the flags. 
We use `<ebay-us-flag>` component for the svg and `<ebay-flag countryCode="us">` for the sprites.
I _believe_ the SVG one will be deprecated and removed because of performance issues. We had those issues when we built phone-input, which is why we swapped to the sprite.

packages/ebayui-core-react/.storybook/preview.jsx
HenriqueLimas: Yep yep! Let me do that! My initial idea was that the current `<ebay-flag>` was easier to do it inline as a span in the phone input and use it instead for the SVG since we do have the same for icon. But since the plan is to replace it with sprite, I will update as you mentioned.

HenriqueLimas: Changed to `EbayFlagsSvg` and we now have `EbayFlag` and `EbayFlagIcon`

packages/ebayui-core-react/src/ebay-flag/__tests__/constants.ts
```diff
@@ -0,0 +1,273 @@
+/* This file is autogenerated, do not edit - it will be overwritten anyway */
+
+import { Icon } from "../types";
+
+export const icons: Icon[] = [
```
agliga, line 5: Can we simply import the other file here rather than have another copy

packages/ebayui-core-react/src/ebay-flag/types.ts
```diff
@@ -0,0 +1,270 @@
+/* This file is autogenerated, do not edit - it will be overwritten anyway */
+
+export type Icon =
```
agliga, line 3: Would be good to have this in a common place so both marko and react can use it.
Anyway can do this in a follow up.

packages/ebayui-core-react/src/ebay-flag/types.ts
```diff
@@ -0,0 +1,270 @@
+/* This file is autogenerated, do not edit - it will be overwritten anyway */
```
agliga, line 1: Is this being generated? I remember we had a script to generate ours.
That said if we move these in a common place then it can be fine

agliga (commented):
Overall LGTM
I'm happy to approve. Just one comment about importing the generated file in another.

packages/ebayui-core-react/src/ebay-flag/__tests__/constants.ts
HenriqueLimas: The other file is type definition. This one is execution and used for adding all the stories. They are both generated

packages/ebayui-core-react/src/ebay-flag/types.ts
HenriqueLimas: It is auto generated yes. We have a script (updated in this PR) to generate it

agliga (approved):

PR #194, opened 2025-06-24
feat: added badge, ccd, cta-button, details, eek, icon-btn in marko 6

agliga:
Also renamed all icons to be `evo-icon-{name}`
Also has 3dplayer and progress spinner changes.


packages/evo-marko/src/tags/evo-badge/index.marko
```diff
@@ -0,0 +1,27 @@
...
+    class: inputClass,
+    type,
+    number,
+    // ...htmlInput
+}=input/>
+
+<const/htmlInput=(({ class: inputClass, type, number, ...rest }) => rest)(
+    input,
+)/>
+<const/roundedNumber=Math.round(+(number === undefined ? NaN : number))/>
```
LuLaValva, line 17: I tested in node, chrome, Safari, and firefox— `Math.round` accepts strings and undefined. We can probably just do
```suggestion
<const/roundedNumber=Math.round(number as number)/>
```

packages/evo-marko/src/tags/evo-ccd/index.marko
```diff
@@ -0,0 +1,58 @@
...
+    min?: string;
+    chargerIcon?: "included" | "not-included";
+    units?: AttrString;
+    /**
+     * The aria-label accessibility label for the ccd component.
+     * This is for internationalization.
+     * It should use min, max, and charger included or not included, and secondaryText in the label in order to demonstrate
+     * to screen readers the content on the component.
+     * Expected return `Charger included. ${min} - ${max} Watts. USB PD`
+     */
```
LuLaValva, line 13: IntelliSense displays this as Markdown, so if your intention is to have multiple paragraphs instead of one big block of text they need to be separated by a blank line

packages/evo-marko/src/tags/evo-ccd/ccd.stories.ts
```diff
@@ -0,0 +1,59 @@
...
+  parameters: {
+    docs: {
+      description: {
+        component: Readme,
+      },
+    },
+  },
+
+  argTypes: {
+    max: {
```
LuLaValva, line 19: We should add `aria-label` here to be explicit

packages/evo-marko/src/tags/evo-details/index.marko
```diff
@@ -0,0 +1,52 @@
...
+    summary,
+    content,
+    as: inputAs,
+    ...rest
+}) => rest)(input)/>
+
+<details
+    ...htmlInput
+    class=["details", inputClass]
+    open:=input.open
```
LuLaValva, line 33: The controllable attributes (in this case `openChange`) _can_ be spread, so it's actually completely okay to just remove this line and `open` from the destructure and it will still be controllable

packages/evo-marko/src/tags/evo-eek/index.marko
```diff
@@ -0,0 +1,89 @@
...
+export interface Input extends Marko.HTML.Div {
+  max: string;
+  min: string;
+  rating: string;
+  /**
+   *  The aria-label accessibility label for the eek rating. This is for internationalization.
+   *  It should use eekRating, min, and max values in the label in order to demonstrate
+   *  to screen readers the content on the component.
+   *  Expected return `Energy Rating: ${eekRating}. Range: ${max} - ${min}.`
+   **/
```
LuLaValva, line 11: Again, this is probably intended to be multiple paragraphs which should be separated by a blank line

packages/evo-marko/src/tags/evo-eek/index.marko
```diff
@@ -0,0 +1,89 @@
...
+<const/htmlInput=(({
+  class: inputClass,
+  size,
+  max,
+  min,
+  rating,
+  ...rest
+}) => rest)(input)/>
+
+<const/eekRating=eekUtil(input)/>
```
LuLaValva, line 63: Not a huge deal, but passing in all of `input` like this is a deopt so Marko will serialize the whole thing instead of just the important parts. I think `eekUtil` should take `min`, `max`, and `rating` as params which are passed individually

packages/evo-marko/src/tags/evo-icon-button/index.marko
```diff
@@ -0,0 +1,68 @@
...
+
+export interface Input extends Omit<Marko.HTML.Button, "type">, Omit<Marko.HTML.A, "type" |`on${string}`> {
+  badge?: Marko.AttrTag<BadgeInput>;
+  badgeNumber?: number | string;
+  href?: string;
+  transparent?: boolean;
+  priority?: "primary" | "secondary" | "tertiary" | "none";
+  size?: "small" | "large";
+  partiallyDisabled?: ButtonInput["partiallyDisabled"];
+  onEscape?: ButtonInput["onEscape"];
```
LuLaValva, line 15: You omitted `"type"` from _both_ `button` and `a`, so it was lost. One option is to add it back like this (I'm not sure why TS didn't just do this by default with no `Omit`)


```
  type?: Marko.HTML.Button["type"] & Marko.HTML.A["type"]
```

packages/evo-marko/src/tags/evo-icon-button/index.marko
```diff
@@ -0,0 +1,68 @@
+static const VALID_SIZES = ["small", "large"];
+import type { AttrString } from "marko/tags-html";
+static var validPriorities = ["primary", "secondary", "tertiary"];
+import type { Input as ButtonInput } from "../evo-button/index.marko";
```
LuLaValva, line 4: It doesn't look like we're actually _using_ `evo-button` here so it doesn't feel right to be coupling the types like this... I think if it were up to me I'd just duplicate the `partiallyDisabled` and `onEscape` types

packages/evo-marko/src/tags/evo-icon-button/index.marko
```diff
@@ -0,0 +1,68 @@
...
+import type { Input as BadgeInput } from "../evo-badge/index.marko";
```
LuLaValva, line 5: We've got a shorthand for this!

```suggestion
import type { Input as BadgeInput } from "<evo-badge>";
```

packages/evo-marko/src/tags/evo-icon-button/index.marko
agliga: I initially was going to extend Icon button from button, but there were some attributes we werent using. 
I think its fine as is. I did add priority and size as well

packages/evo-marko/src/tags/evo-icon-button/index.marko
```diff
@@ -0,0 +1,69 @@
+static const VALID_SIZES = ["small", "large"];
+import type { AttrString } from "marko/tags-html";
+static var validPriorities = ["primary", "secondary", "tertiary"];
```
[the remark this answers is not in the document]
agliga: Actually, going to just import that from button

LuLaValva (approved):

PR #241, opened 2025-07-30
feat(card): added new component

### LINKED ISSUE #74: card: marko component
### Description

ebayui 

### Screenshots

_No response_

### Figma link

_No response_

- [ ] Successful design review with Design System Team.
- [ ] Supports dark mode.
- [ ] Successful accessibility review with Core A11Y Team.
- [ ] Successful frontend review with eBayUI Team.

agliga:
Fixes #74

## Description
* Added new card component
* Added new marko 6 component (note, tests in marko 6 are not being run yet)

## Checklist

### DIFF
--- a/packages/ebayui-core/src/components/ebay-card/examples/anchor.marko
+++ b/packages/ebayui-core/src/components/ebay-card/examples/anchor.marko
@@ -1,0 +1,19 @@
+class {}
+<ebay-card
+    on-click("emit", "click")
+    href="https://www.ebay.com"
+    file={
+        name: "file-name.jpg",
+        type: "image/jpeg",
+        src: "https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg",
+    }
+    ...input
+>
+    <@overline text="Authenticity Guaranteed">
+        <ebay-authenticity-guarantee-16-icon/>
+    </@overline>
+    <@title>Authentic Rookie Cards</@title>
+    <@description>
+        Professional-grade rookie cards that appreciate faster than most people's cooking skills and last longer than New Year's resolutions.
+    </@description>
+</ebay-card>
--- a/packages/ebayui-core/src/components/ebay-card/examples/anchor.marko
+++ b/packages/ebayui-core/src/components/ebay-card/examples/anchor.marko
@@ -2,15 +2,12 @@ class {}
 <ebay-card
     on-click("emit", "click")
     href="https://www.ebay.com"
-    file={
-        name: "file-name.jpg",
-        type: "image/jpeg",
-        src: "https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg",
-    }
     ...input
 >
-    <@overline text="Authenticity Guaranteed">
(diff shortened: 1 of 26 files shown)


packages/evo-marko/src/tags/evo-card/index.marko
```diff
@@ -0,0 +1,80 @@
...
+        layout === "horizontal" && "card--horizontal",
+        inputClass,
+    ]
+    disabled=elem === "button" && disabled
+    type=elem === "button" && "button"
+    href=elem === "a" && (disabled ? "" : href)
+    tabindex=disabled && -1
+    ...htmlInput>
+    <div class="card__media">
+        <img ...file>
```
LuLaValva, line 55: What happens here if this is actually a [File](https://developer.mozilla.org/en-US/docs/Web/API/File/File), as suggested in the `Input` type? I don't think it makes sense to spread that into ``

LuLaValva (commented):
Just one thing I'm confused about, but otherwise looks good

packages/ebayui-core/src/components/ebay-card/index.marko
```diff
@@ -0,0 +1,94 @@
...
+    ]
+    disabled=elem === "button" && disabled
+    on-click(!disabled && "emit", "click")
+    type=elem === "button" && "button"
+    href=elem === "a" && (disabled ? "" : href)
+    tabindex=disabled && -1
+    ...processHtmlAttributes(htmlInput)
+>
+    <div class="card__media">
+        <img ...processHtmlAttributes(file || [])>
```
HenriqueLimas, line 57: `file` is not a required type, should we render `card__media` only when `file` is present?

packages/evo-marko/src/tags/evo-card/index.marko
HenriqueLimas: Should we do something similar that we do on [file-preview-card](https://github.com/eBay/evo-web/blob/main/packages/ebayui-core/src/components/ebay-file-preview-card/index.marko#L44-L59) where we check for the type and render accordingly?

packages/ebayui-core/src/components/ebay-card/examples/anchor.marko
```diff
@@ -0,0 +1,19 @@
...
+    on-click("emit", "click")
+    href="https://www.ebay.com"
+    file={
+        name: "file-name.jpg",
+        type: "image/jpeg",
+        src: "https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg",
+    }
+    ...input
+>
+    <@overline text="Authenticity Guaranteed">
```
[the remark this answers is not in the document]
agliga: I think the reason why this is like this is because there was a `card__overline-text` class being applied.
When I saw it initially, I thought this had styles associated with it. But looking again, there are no styles associated with it, so I will do this suggestion.

packages/ebayui-core/src/components/ebay-card/index.marko
agliga: I will swap this to image attribute tag. I think its cleaner

agliga:
I swapped out of using file to just using `@image` directly like I did with the chips PR
Also addressed all comments. 

I was initially going to use file preview card but there were some issues integrating with it, so that was left over.

LuLaValva (approved):

PR #257, opened 2025-08-18
feat(section-notice): update cta without footer to latest specs

### LINKED ISSUE #18: notice: rework footer and cta sections
CTA should be positioned under body on small screen in all cases, even when there is no dismiss button:
Screenshot 2023-09-01 at 11 30 19 AM

We can see that when there is a CTA with dismiss button, the CTA is positioned correctly.

Screenshot 2023-09-01 at 11 26 41 AM

When we replace the dismiss button with the CTA, the CTA loses the ability to go in the correct space under the body:
Screenshot 2023-09-01 at 11 26 30 AM

Even if trying to use just the `section-notice__cta` (not sure if this is the desired expectation - it is not clear from current docs or stories), there is a spacing issue on large screens:

Screenshot 2023-09-01 at 11 36 57 AM

This behavior can be verified in current storybook: https://opensource.ebay.com/skin/storybook/?path=/story/skin-section-notice--information-with-button

darkwebdev:
Partially fixes #18 spacing on the desktop issue.

- [x] This PR contains CSS changes
- [ ] This PR does not contain CSS changes

## Description
The current implementation of section-notice with CTA has padding to the right which is not needed on desktop. We removed the padding and created a new Section Notice story called Link CTA inside CTA Section for this.

## Notes
This is only applied to CTA without dismiss button.

### dWeb
image

### dWeb RTL
image

### mWeb
image

### mWeb RTL
image

## Checklist

- [x] I verify the build is in a non-broken state

- [x] I regenerated all CSS files under dist folder
- [x] I added/updated/removed Storybook coverage as appropriate

### DIFF
--- a/.changeset/large-mice-wink.md
+++ b/.changeset/large-mice-wink.md
@@ -1,0 +1,5 @@
+---
+"@ebay/skin": patch
+---
+
+feat(section-notice): update cta without footer to latest specs


packages/skin/src/sass/section-notice/section-notice.scss
```diff
@@ -145,9 +145,12 @@ p.section-notice__cta {
...
         justify-self: flex-end;
-        margin-right: var(--spacing-200);
         margin-top: 0;
-        padding-right: var(--spacing-200);
+
+        .section-notice:has(.section-notice__dismiss) & {
+            margin-inline-end: var(--spacing-200);
+            padding-inline-end: var(--spacing-200);
+        }
     }
```
agliga, line 154: So while this works, couple of concerns
1. First, we dont nest things even though it works and is less code, its definitly harder to read. 
2. It seems like a big code smell to change these margins based on the presense of another element.

I looked a bit locally and found a probably better solution. 
Currently we are setting the footer element to be grid column 4 but we also set the CTA to be column 4 (which doesn't seem to make sense). To fix that, we apply some margin to position it so it doesn't overlap. 
However, what is probably a better fix is to set CTA to be column 3 and leave footer as 4. 
This requires one more fix to change the columns to be `min-content` on the section notice.
`grid-template-columns: 32px auto min-content min-content;`
Playing around with that, it works better and seems to fix this issue much cleaner. 

Let me know if you can try that solution and if you run into any other issues when you do it.

agliga (commented):
When I looked at the PR initially I didn't dive in but I had thought we could add the margin/paddings to the footer element instead of the CTA, but trying it locally I saw yeah, that doesn't work. 
So I offer another solution which is to fix the whole grid issues in this component, by just using the right grid layout instead of adding arbitrary padding/margins.

agliga:
Just checking if there are any updates planned on this PR or not? Will close if not!

darkwebdev:
@agliga 
> Just checking if there are any updates planned on this PR or not? Will close if not!

I've fixed the PR, please take a look

ArtBlue:
> @agliga
> 
> > Just checking if there are any updates planned on this PR or not? Will close if not!
> 
> I've fixed the PR, please take a look

@darkwebdev , can you also respond to the Slack thread about why this is needed? You were tagged a few days ago.

agliga (commented):
Overall LGTM. I feel its much more robust now.
Lets just double check the playbook versions of what we have and update it accordingly. 
The versions should be in the page.marko in section notice. 
Im happy to approve it after that.

agliga (dismissed):

PR #261, opened 2025-08-20
refactor(ts): switch from private Attr types to public Marko.HTMLAttributes

LuLaValva:
## Description

- The `Attr*` types were supposed to be private, but I was using them freely throughout the eBayUI types
- While fixing these types, I ran into a few minor type bugs and fixed them. For some of these fixes, a minor refactor was required


packages/ebayui-core/src/components/ebay-chips-combobox/index.marko
```diff
@@ -4,7 +4,6 @@ $ const {
     class: inputClass,
     fluid,
     error,
-    value,
```
LuLaValva, line 7: This was unused, so I removed it

packages/ebayui-core/src/components/ebay-date-textbox/component.ts
```diff
@@ -16,28 +16,15 @@ export interface InvalidDateEvent {
     index: number;
 }
 
-interface DateTextboxInput {
+interface DateTextboxInput extends Omit<CalendarInput, `on${string}`> {
```
LuLaValva, line 19: Most of these types can be removed if we extend calendar, since they are directly piped through

packages/ebayui-core/src/components/ebay-phone-input/component.ts
```diff
@@ -15,14 +15,14 @@ export interface PhoneInputEvent {
     countryCode?: string;
 }
 
-interface PhoneInputInput extends Omit<ListboxButton, `on${string}`> {
+interface PhoneInputInput extends Omit<Marko.HTML.Span, `on${string}`> {
```
LuLaValva, line 18: This type was wrong, the spread goes into a `<span>` and _some_ attributes are piped into `ListboxButton`

packages/evo-marko/src/tags/evo-ccd/index.marko
```diff
@@ -52,7 +53,7 @@ export interface Input extends Marko.HTML.Div {
             <evo-icon-ccd-top class="ccd__top-icon"/>
             <div class="ccd__body">
                 <div>${min} - ${max}</div>
-                <div>W</div>
+                <div>${units}</div>
```
LuLaValva, line 56: Seems like `<evo-ccd>` doesn't have all features from `<ebay-ccd>`, I went ahead and added this `units` since I didn't want to keep an unused type but am happy to revert

LuLaValva (commented):
Did a self review to mark changes that might need a closer look

agliga (approved):
Thanks for this!

PR #274, opened 2025-08-26
feat(marko6): added skeleton

saiponnada:
- Added Skeleton


packages/evo-marko/src/tags/evo-skeleton/test/__snapshots__/test.server.ts.snap
```diff
@@ -0,0 +1,80 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
agliga, line 1: I see the snapshots, but I don't see the tests
Lets copy them over just so we don't forget them. (and rename them to typescript files)

agliga (changes requested):
LGTM, only thing is if we can make sure the tests are copied over (even if they aren't working 100%)

PR #275, opened 2025-08-29
feat(textbox): added marko6 component

agliga:
Marko 6 components of textbox and filter-input

There is currently a bug with events on dynamic tags in marko 6, but the floating label and clear button do work if the textbox is changed from a dynamic tag to an input.


LuLaValva (commented):
One round of pedantic comments. Looks great overall!

LuLaValva (approved):

PR #294, opened 2025-09-19
feat: roving tabindex and tabs component

agliga:
WIP, not fully working
Added tabs and roving tabindex makeup component


agliga:
> Just checking, this doesn't actually use the makeup-roving-tabindex module though, does it?

It does not. Its a marko version of makeup roving tabindex.

PR #295, opened 2025-09-19
feat: evo-accordion

saiponnada:
Marko 6 version of accordion


.changeset/fast-parts-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@evo-web/marko": patch
```
agliga, line 2: No need for changeset

packages/evo-marko/src/tags/evo-accordion/examples/autoCollapsed.marko
```diff
@@ -0,0 +1,20 @@
+<evo-accordion ...input autoCollapse=true>
```
agliga, line 1: Lets change these examples to use the new API for showing that it works. 
Basically `autoCollapse` should not be used anymore
We should pass in `open=-1` which will treat it as auto collapse=true

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
+import type { Input as DetailsInput } from "<evo-details>";
+export interface Input extends Omit<Marko.HTML.UL, `on${string}`> {
+    size?: "regular" | "large";
+    autoCollapse?: boolean;
```
agliga, line 4: This should not be needed anymore.

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
...
+    ...htmlInput>
+    <for|item, index| of=items>
+        <li>
+            <evo-details
+                ...item
+                open=(
+                    Array.isArray(open) ? open.includes(index) : open === index
+                )
+                openChange(isOpen) {
+                    if (autoCollapse) {
```
agliga, line 39: This test should be 
```
if (!Array.isArray(open))
```
So if they pass in `open` as an array, treat it as a multi select and they can have multiple sections open. 
If they pass in `open` as a number, then only 1 section can be opened at a time (IE autoCollapse is enabled)

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
+import type { Input as DetailsInput } from "<evo-details>";
+export interface Input extends Omit<Marko.HTML.UL, `on${string}`> {
```
agliga, line 2: We will need to add a generic here because that way the type will be inferred automatically in the parent component
Also remove the `Omit`, its no longer needed

So just do this

```
export interface Input<Open extends number | number[] | undefined> extends Marko.HTML.UL {
   // other defitnitions
   open: Open;
   openChange: (open: Open) => void;
}
```

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
...
+                            open = index;
+                        } else {
+                            open = -1;
+                        }
+                    } else {
+                        const openArray = Array.isArray(open)
+                            ? open
+                            : typeof open === "number" && open >= 0
+                              ? [open]
+                              : [];
```
agliga, line 50: This will not be needed anymore if the original if statement checks if its an array or not.

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
...
+                        if (!isOpen) {
+                            open = openArray.filter((n) => n !== index);
+                        } else if (!openArray.includes(index)) {
+                            open = [...openArray, index];
+                        }
+                    }
+
+                    if (open !== undefined) {
+                        input.openChange?.(open);
+                    }
```
agliga, line 61: Not needed!
When you set 
`<let/open:=input.open>`
Then whenever you set open like `open = newOpenVariable` it atomically calls `openChange`

packages/evo-marko/src/tags/evo-accordion/index.marko
```diff
@@ -0,0 +1,67 @@
...
+                }>
+                <${item.content}/>
```
agliga, line 63: This is also not needed, if you spread in `...item` then `item.content` is also spread in. Double check, but removing this should be safe to do!

agliga (commented):
Good second pass, thanks for updating, but I think there can be a bit more clean up! Let me know if you want to go over any of these offline!

Issue #112, opened 2023-06-28
[react] Remove legacy React APIs usage


darkwebdev:
https://react.dev/reference/react/legacy

Legacy React APIs

These APIs are exported from the react package, but they are not recommended for use in newly written code. See the linked individual API pages for the suggested alternatives.
Legacy APIs

- **Found in 26 files**: [Children](https://react.dev/reference/react/Children) lets you manipulate and transform the JSX received as the children prop. [See alternatives.](https://react.dev/reference/react/Children#alternatives)
- **Found in 20 files**: [cloneElement](https://react.dev/reference/react/cloneElement) lets you create a React element using another element as a starting point. [See alternatives.](https://react.dev/reference/react/cloneElement#alternatives)
- **Found in 2 files**: [createElement](https://react.dev/reference/react/createElement) lets you create a React element. Typically, you’ll use JSX instead.
- **Found in EbayPagination** and tests/stories: [createRef](https://react.dev/reference/react/createRef) creates a ref object which can contain arbitrary value. [See alternatives.](https://react.dev/reference/react/createRef#alternatives)
- Not found: [Component](https://react.dev/reference/react/Component) lets you define a React component as a JavaScript class. [See alternatives.](https://react.dev/reference/react/Component#alternatives)
- Not found: [isValidElement](https://react.dev/reference/react/isValidElement) checks whether a value is a React element. Typically used with [cloneElement.](https://react.dev/reference/react/cloneElement)
- Not found: [PureComponent](https://react.dev/reference/react/PureComponent) is similar to [Component,](https://react.dev/reference/react/Component) but it skip re-renders with same props. [See alternatives.](https://react.dev/reference/react/PureComponent#alternatives)

ArtBlue:
This is part of the new evo component. Closing per team.

PR #304, opened 2025-09-30
chore(EbayAvatar): remove usage of React.Children and cloneElement

HenriqueLimas:
Part of #112 

## Description

- Remove `findComponents` usage from `EbayAvatar`
- Introduce a context so state update function can be used on the `EbayAvatarImage`
- Move onLoad logic to `EbayAvatarImage` component

## Notes

This doesn't change any behavior nor API of the component.


agliga (approved):
LGTM

PR #305, opened 2025-10-01
feat(listbox-button): added new marko 6 component

agliga:


packages/evo-marko/src/tags/evo-listbox-button/examples/with-description.marko
```diff
@@ -0,0 +1,14 @@
+<evo-listbox-button prefixLabel="Select" name="formFieldName" ...input>
+    <@option selected value="1">
```
LuLaValva, line 2: Same goes for this one

LuLaValva (commented):
Just a few nitpicks, but otherwise looks great! These are looking better and better

PR #343, opened 2025-10-30
fix(text-box): correct label width calculation for textbox

dukchen:
## Description

Fixed the floating label width calculation for textbox components to properly account for padding, border, and scale transformations.

## Notes

- Updated `floating-label.scss`
- Changed label width from `calc(100% - 40px)` to `calc((100% - var(--spacing-200) * 2 - 2px) / 0.75)`
- The new calculation properly accounts for:
    - Padding: `var(--spacing-200) * 2` (left and right padding from textbox control)
    - Border: `2px` (textbox border 1px left + 1px right)
    - Scale factor: `0.75` (the label is scaled down to 75%, so we divide by 0.75 to get 100% width)

## Problem Solved

Previously, the label width was set to a fixed `100% - 40px`, which didn't accurately reflect the actual available space for the label. This caused the label to not properly fill the available width, especially when considering the textbox's padding and border, and the label's scale transformation.

## Screenshots

Before
Screenshot 2025-10-30 at 12 16 10

After
Screenshot 2025-10-30 at 12 15 03

## Checklist

### DIFF
--- a/.changeset/khaki-ravens-hear.md
+++ b/.changeset/khaki-ravens-hear.md
@@ -1,5 +1,5 @@
 ---
-"@ebay/skin": major
+"@ebay/skin": patch
 ---
 
 Fixed floating label width calculation to account for 0.75 scale transformation and added width transition animation to prevent labels from


agliga:
I think the main thing needed here is to make sure the text isnt shooting out of the label when it animates. 
Otherwise it looks fine

PR #346, opened 2025-11-01
refactor: share code for components when possible

LuLaValva:
This proposes some ideas for sharing code between components in the evo site. It has a few issues which I'll talk to Marko team and @ArtBlue about, but in general I think this will make components much easier to add and work on.

## Important Files

Since I needed to move files between directories, there are lots of files changed. _Most_ can be ignored as they don't have code changes. For the review here, please look at

- Modified/removed files in `accordion/`
- New files in `$name/`


src/routes/component/_tabs/+layout.marko
```diff
@@ -0,0 +1,8 @@
+// pulling the name off of the route, because Marko Run doesn't
+// have a way to share dynamic and static directories yet
+$ const pathSegments = $global.url.pathname.split("/")
+$ const componentName = pathSegments[pathSegments.findIndex(v => v === "component") + 1]
```
agliga: Is there a way we can start writing components in marko 6?

src/routes/component/_tabs/$name/react+page.marko
ArtBlue: This path needs a dedicated `meta.json` since it needs its own distinct identity. We can't have duplicate titles and meta descriptions.

src/routes/component/_tabs/accordion/css+page.marko
ArtBlue: Would this specific page have a dedicated meta tags if we create a `css+meta.json` here?

agliga:
Closing for now

PR #384, opened 2025-12-02
fix(ebaynumberinput): Styling fix for label order for EbayNumberInput

bpushpkar:
- Fixes #
Before:
Screenshot 2025-12-02 at 11 28 25 AM

After:
Screenshot 2025-11-28 at 1 06 33 PM

## Description

- Added order for label, and icon buttons, increment, decrement, input for EbayNumberInput.

## Checklist

### DIFF
--- a/packages/skin/dist/number-input/number-input.css
+++ b/packages/skin/dist/number-input/number-input.css
@@ -1,5 +1,5 @@
 @keyframes number-input-decrement-before {
-    from {
+    0% {
         transform: translateY(0);
     }
     to {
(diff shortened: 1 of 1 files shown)


agliga (commented):
I believe this is not a change we want in number input 
What issue is this trying to solve? We require an issue or communication with our team before any fix.
I don't see any issues associated with this component. Also this component is also coded as per designs and shouldn't need any modifications

agliga:
But that's not a valid usecase. The delete button should not be used in the label usecase but only without a label.

PR #414, opened 2026-01-07
feat(ci): integrate Percy visual regression testing into CI/CD

HenriqueLimas:
## Description

Add automated Percy visual regression testing to GitHub Actions workflow with partial snapshot detection for faster builds.

Features:
- Automatic component change detection from git diff
- Partial snapshots for changed components only based on a map of dependencies
- Full snapshots for global changes (tokens, variables)
- Added Percy application integration
- Full snapshots on main and Auto-approval
- Path filtering to skip non-Skin changes

Changes:
- Add Percy GitHub Actions workflow (.github/workflows/percy.yml)
- Add Github action to detect module that changed (.github/actions/detect-changed-components)
- Update CONTRIBUTING.md with automated workflow documentation
- Add PERCY-FAQ.md for comprehensive user guidance

Example PR: https://github.com/eBay/evo-web/pull/430

### DIFF
--- a/.github/actions/detect-changed-components/index.mts
+++ b/.github/actions/detect-changed-components/index.mts
@@ -83,6 +83,19 @@ async function extractStoryTitle(filePath: string): Promise<string | null> {
 }
 
 /**
+ * Extract component name from story title
+ * @param title - Story title (e.g., "Skin/Alert Dialog" or "Skin/Button/Primary")
+ * @returns Component name (e.g., "Alert Dialog" or "Button") or null if invalid format
+ */
+function extractComponentNameFromStoryTitle(title: string): string | null {
+  const parts = title.split("/");
+  if (parts.length >= 2 && parts[0] === "Skin") {
+    return parts[1];
+  }
+  return null;
+}
+
+/**
  * Get all story titles for a given component directory
  * @param componentDir - Component directory name (e.g., 'alert-dialog')
  * @param sassDir - Path to SASS directory
@@ -103,11 +116,9 @@ async function getStoryTitlesForComponent(
     const title = await extractStoryTitle(absolutePath);
 
     if (title) {
-      // Extract just the component part: "Skin/Alert Dialog" → "Alert Dialog"
-      // "Skin/Button/Primary" → "Button" (parent category only)
-      const parts = title.split("/");
-      if (parts.length >= 2 && parts[0] === "Skin") {
-        componentNames.add(parts[1]); // Get the component name after "Skin/"
+      const componentName = extractComponentNameFromStoryTitle(title);
+      if (componentName) {
+        componentNames.add(componentName);
       }
     }
   }
@@ -236,10 +247,8 @@ async function main(): Promise<void> {
         const title = await extractStoryTitle(absolutePath);
 
         if (title) {
-          // Extract component name: "Skin/Button/Base" → "Button"
-          const parts = title.split("/");
-          if (parts.length >= 2 && parts[0] === "Skin") {
-            const componentName = parts[1];
+          const componentName = extractComponentNameFromStoryTitle(title);
+          if (componentName) {
             core.info(`  ${file} → ${componentName} (from story)`);
             directStoryTitles.add(componentName);
           }


HenriqueLimas:
Right now the story detection script is trying to map the the module that changed to a story with the name, but not all the stories maps the same name as the module. I am updating to search for the stories instead to be more precise

packages/skin/gulpfile.js
```diff
@@ -1,14 +1,15 @@
 var child_process = require("child_process");
 var yargs = require("yargs");
+const { hideBin } = require("yargs/helpers");
 
 async function runSnapshots(storiesList, isDryRun) {
```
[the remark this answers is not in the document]
HenriqueLimas: Removed

packages/skin/scripts/detect-changed-components.js
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env node
```
agliga: Can we have all the scripts be in typescript? We did the conversion to all typescript scripts recently. 
We can use `tsx` to run the scripts

packages/skin/scripts/detect-changed-components.js
```diff
@@ -0,0 +1,240 @@
...
+        // Check for global changes first (existing behavior)
+        if (hasGlobalChanges(changedFiles)) {
+            console.log("all");
+            return;
+        }
+
+        // NEW: Build dependency graph and find all affected files
+        let allAffectedFiles = new Set();
+
+        try {
```
agliga: This seems overly complex. 
So I would think this should be the path forward
1. Get the list of changed files (`getChangedFiles()`, only need to get the actual title, like `button`)
2. Remove the ignored directories
3. Add dependencies in the list of changed files (just use the `config.ts` I mentioned earlier, its a simple key lookup)
4. Convert all those names from camel case to proper case. 
5. Done.
That should be done in maximum 10-15 lines.

packages/skin/scripts/percy-wait-and-comment.js
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env node
```
agliga: Is this file only to show the status of the percy command?
Its probably better if we get the github integration working than adding an extra file to maintain.

packages/skin/scripts/sass-dependency-analyzer.js
```diff
@@ -0,0 +1,355 @@
...
+ * @returns {Promise<Map<string, Array<string>>>} Promise resolving to map of file path → array of dependency paths
+ *
+ * @example
+ * const graph = await buildDependencyGraph('/path/to/sass');
+ * // Returns: Map {
+ * //   '/path/to/button/button.scss' => ['/path/to/variables/variables.scss', ...],
+ * //   ...
+ * // }
+ */
+async function buildDependencyGraph(sassDir) {
```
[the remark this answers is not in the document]
HenriqueLimas: Here we are checking for mixins. So if a mixin changes we are filtering which files the mixins is being used to trigger the stories for those components. Also, It doesn't seem the `addModules` is being used anywhere today, is that up to date? But that would be a good point, we need to also check these inter dependencies, right now its not, I will think about this case and how we can solve them.

packages/skin/scripts/detect-changed-components.js
HenriqueLimas: This approach will miss the mixin changes and any other @use/@forward usage. I do agree that we still need the mapping (adding it in a separate commit). Also I was doing a simple conversion from camel to proper case before, but some story names have different mapping so it might miss some stories or fail, so I updated to search for the actual story names to be more precise.

.github/actions/detect-changed-components/sass-dependency-analyzer.js
```diff
@@ -0,0 +1,387 @@
...
+ * Module dependencies map
+ *
+ * Maps: changed module → modules that need to be re-snapshotted
+ * These represent CSS class-based dependencies found in story files.
+ *
+ * For example, the calendar module uses icon-button class in its stories,
+ * so when icon-button changes, calendar needs Percy snapshots even though
+ * there's no Sass @use/@forward relationship.
+ */
+const MODULE_DEPENDENCIES = {
```
agliga: Ideally I would like this to be in 1 place so we don't have to change it multiple times.
If its inside a file like this it might be hard to find. 
We can put it in the root as a config.json file if needed and have both the module builder as well as this use it. 
We can also put it in component-metadata, which might be better (like adding a dependency key to it)

agliga (commented):
Its looking better now. 
A couple of minor comments:
1. I think we should be relying on dist changes as well. We have times where we have no src changes but dist changes due to new build steps. 
2. Im almost of the opinion to put mixins in the global check. Mixins don't change too often and there seems to be a lot of code to just handle the mixins. Not saying its bad but it adds extra maintinance. We could remove the depedency analysis file if we add mixins to the global path check
3. Last thing in the global files changed, again we might want to use dist there. For example our token src files might not change very much but the dist will change when a new token package is added.

HenriqueLimas:
@agliga agree, I will simplify to consider mixins as global. Also, we already have a `submodules` field, so I will use that one instead of adding a new one and duplicate it. In case it is needs, I will update any dependency that is needed. I will also double check the dist file. Because maybe we should only check the `dist` folder.

ArtBlue (dismissed):
This looks good from my perspective. Just one minor comment on dry runs.

docs/adr/2026-01-16-percy-ci-automation.md
```diff
@@ -0,0 +1,110 @@
+# Percy Snapshots CI/CD Automation
```
agliga: Lets do this in a follow up PR. It seems good to have this but this is too verbose for what we need.

.github/actions/detect-changed-components/index.js
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * Extract the actual story title by importing the .stories.js file
+ * @param {string} filePath - Absolute path to .stories.js file
+ * @returns {string|null} Story title (e.g., "Skin/Alert Dialog") or null
+ */
+function extractStoryTitle(filePath) {
+  try {
+    // Import the story file
+    const storyModule = require(filePath);
```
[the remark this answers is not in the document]
HenriqueLimas: I will check if we can use typescript with github action and use esm too.

PR #454, opened 2026-01-23
feat(tests): unit and interaction tests for button component in marko and react 

reidmore-online:
## Description

This PR adds unit and interactions tests for the ebay-button components in ebayui-core and ebayui-core-react. It also adds a folder and files for AI context for Claude and Cline. 

The test files are mostly unaltered and may not all pass, many are possibly not needed, so feel free to suggest tests to remove or modify!


packages/ebayui-core/src/components/ebay-button/button.stories.ts
```diff
@@ -222,6 +223,320 @@ Default.parameters = {
...
 
+Default.play = async ({ canvasElement, step }: { canvasElement: HTMLElement; step: any }) => {
+    const canvas = within(canvasElement);
+    
+    const button = canvas.getByRole("button");
+
+    await step("Check initial rendering", async () => {
+        await expect(button).toBeInTheDocument();
+        await expect(button).not.toBeDisabled();
+        await expect(button.textContent?.trim()).toBe("Button");
```
saiponnada: We need to set rules on what are the stuff we want in unit vs interaction tests to be. If not these are repetitive and increases our time to run these. For e.g  `await expect(button.textContent?.trim()).toBe("Button");` or large button might not be that useful IMO. A good example for the button is loading a button with loading state and then showing an actual button shows the depth of interaction.

agliga (commented):
I think there's a lot of good here
But I also think there's a lot of extra things that we don't need or we are already testing. 

The good:
- Storybook interaction tests are great and neat
- accessibility vitest focus testing and keyboard testing is great

The could be better:
- We should really have a common place for the accessbility storybook interactions which should be framework neutral and can be run by both marko and react (I added a comment with suggestions on how to get that, Im sure adding that in AI it could do it!)
- As commented, theres a lot of tests we are already test. While of course, running these shouldn't be adding much to it, it also increases maintenance costs. We used to have a lot of these testing presence of certain attributes but we got rid of them and went to snapshot tests which capture all that info and make it easy to see what changed. If we want to keep those interactions I would make them take a snapshot of the HTML and store it in order to test rather than checking each and every attribute. 

Again, we can move forward if needed but these are long term things that I would like addressed before we do.

docs/ai/agents.md
ArtBlue: Would all models automatically use this `agents.md`?

docs/ai/testing.md
ArtBlue: Is this kind of like a skill? How does the agent follow these instructions?

reidmore-online:
@ArtBlue I could remove the context files if that makes sense, I'm more concerned about the tests themselves (which were generated using the files).

PR #460, opened 2026-01-26
refactor: use define for evo-card

LuLaValva:
Use a `<define>` pattern for switching between tags, instead of a dynamic tag with a `<const>`


LuLaValva:
Good idea @ianmcburnie! I don't think it makes sense to put in a style guide for this repo, but it's a cool thing to put in some kind of "recommended patterns" doc on the Marko site

PR #474, opened 2026-01-28
docs: Added first ADR

ianmcburnie:
## Description

Adds our first ADR, which is the decision to use ADRs!

## Notes

Much of the what/why/how was thrashed out "offline" and this was the decision made.

### DIFF
--- a/docs/adr/0001-adr-template.md
+++ b/docs/adr/0001-adr-template.md
@@ -1,4 +1,4 @@
-# 1. Architecture decisions
+# 1. ADR Template
 
 Date: 2026-01-28
 


ArtBlue:
> > Should add a template that we should use instead? I don't think we need an ADR about using this ADR. If we have a template, then we should simply follow it.
> 
> This is what we did for our internal VAT (our first ADR is saying we will be using ADRs - kinda funny I know - but it just gets it seeded with something I guess and shows the template we should be following).
> 
> @agliga Would renaming it to `0001-decision-format.md` work? If this decision changes, all new adrs from that time forward would adopt the new format.

I think that helps. We could also do `0000-adr-template` to be more explicit. Just a thought as it removes all ambiguity.

agliga (approved):

PR #493, opened 2026-02-04
chore: removed percy from checklist

agliga:
Removed percy from checklist for PR. Since PRs run percy automatically we don't need this checklist item


ianmcburnie (commented):
Percy build needs to be approved in Percy before github merge can happen, right?

agliga:
> Percy build needs to be approved in Percy before github merge can happen, right?

Yes

PR #504, opened 2026-02-10
feat(marko6): added tristate checkbox

agliga:
* I used an array in <let>. Happy to do it differently, but this seems to work the best. This is also probably ideal when you have a list of items and want to submit the form too.


LuLaValva (commented):
We should add a simple `controlled` example like we've done for all other controllable components, and _maybe_ consider adding support for `checkedValue` and `checkedValueChange` (discussed offline)

Issue #435, opened 2026-01-15
dialog: add support for native HTML dialog


agliga:
### Description

We should move all our dialog code in skin to use native html dialog tag. 
This is now supported by our browser policy. 
We should still support the old dialog pattern though. The goal will be on marko 6 components we will use the new dialog pattern. 

### Engineering pre-requisites

- [ ] Successful design review with Design System Team.
- [ ] Supports dark mode.
- [ ] Supports responsive design.
- [ ] Tokens have been provided (or already exist).
- [ ] Successful accessibility review with Core A11Y Team and MIND pattern is created.
- [ ] Successful frontend review with eBayUI Team.

PR #506, opened 2026-02-10
feat(dialog): added new component using base <dialog>

agliga:
* fixes https://github.com/eBay/evo-web/issues/435
## Description
* Added `<dialog>` component. This is the skin/evo-web piece. 
* Used a lot of the language from lightbox dialog and updated it to support `dialog` html element
* No more need of `aria-modal="true"` or `role=dialog`
* Added JS mainly for supporting the animation as well as the open/close for the buttons. This is in a separate file (`dialog.ts`).

## Note
* There is a marko component but that is only there so the code does not get lost. Its not a working solution. That shouldn't affect anything since its on the evo-marko library

### Large
Screenshot 2026-02-19 at 10 06 13 AM

### Small 
The only change to be before is that the small screen was positioned at the bottom. This is positioned in the middle as design prefers it this way. 
Screenshot 2026-02-19 at 10 06 27 AM

### DIFF
--- a/packages/skin/dist/dialog/dialog.css
+++ b/packages/skin/dist/dialog/dialog.css
@@ -63,6 +63,7 @@ body:has(.dialog[open]) {
     background-color: var(--color-background-primary);
     border: none;
     border-radius: var(--border-radius-150);
+    color: var(--color-foreground-primary);
     flex-direction: column;
     max-height: 90%;
     max-width: 88%;
--- a/packages/skin/src/sass/dialog/dialog.scss
+++ b/packages/skin/src/sass/dialog/dialog.scss
@@ -75,6 +75,7 @@ body:has(.dialog[open]) {
     background-color: var(--color-background-primary);
     border: none;
     border-radius: var(--border-radius-150);
+    color: var(--color-foreground-primary);
     flex-direction: column;
     max-height: 90%;
     max-width: 88%;


src/routes/_index/components/dialog/accessibility+meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "pageTitle": "Dialog Accessibility Guidelines — Building Inclusive Image Modals",
+  "pageDescription": "Learn how to make dialog components fully accessible with proper focus trapping, keyboard navigation, and ARIA dialog roles. Follow these WCAG-aligned guidelines to ensure image overlays work for all users with assistive technologies."
```
ianmcburnie: This will need updating to account for the fact we are using native dialog now

src/routes/_index/components/dialog/accessibility+meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "pageTitle": "Dialog Accessibility Guidelines — Building Inclusive Image Modals",
```
ianmcburnie: This doesn't look right.

src/routes/_index/components/dialog/css+meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "pageTitle": "Dialog CSS Component",
+  "pageDescription": "A Dialog CSS Component is a responsive and accessible dialog that is used to display dialogs."
```
ianmcburnie: This doesn't sound right.

src/routes/_index/components/dialog/dialog.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    {
+      once: true,
+    },
+  );
+}
+
+// Find all dialog elements
+for (const dialog of document.querySelectorAll(".dialog")) {
+  dialog.setAttribute("closedby", "any");
```
ianmcburnie: Did you check browser support for `closedby`? It's showing as limited availability in MDN. Or is this a safe progressive enhancement usage?

src/routes/_index/components/dialog/dialog.ts
```diff
@@ -0,0 +1,34 @@
+function closeDialog(dialog: HTMLDialogElement) {
+  dialog.classList.add("dialog--close");
+  dialog.addEventListener(
```
ianmcburnie: Will we have similar JS logic in the Marko component? If so this would be a good candidate to put in our new js utilities (i.e. makeup)

ianmcburnie:
I'd kind of like to see a new vanilla js module/class for this dialog. It can handle the closing animation and closedby logic. e.g. `@import 'dialog' from @evo-web/js/ui`. That will be an easy way to see the gaps we still have to _makeup_ based on browser policy, and provide a single source of truth for those.

src/routes/_index/components/dialog/css+meta.json
ianmcburnie: That sounds much better.

LuLaValva (commented):
Doesn't work in dark mode! Looks like `<dialog>` overrides `color` from its parent

Image

Otherwise it looks good I think. The only potential thought is— do we want to keep this `dialog--close` behavior even though _in theory_ we should be able to move away from it pretty soon and leave close animations as a progressive enhancement for newer browsers?

agliga:
> Doesn't work in dark mode! Looks like `<dialog>` overrides `color` from its parent
> 
> Image
> Otherwise it looks good I think. The only potential thought is— do we want to keep this `dialog--close` behavior even though _in theory_ we should be able to move away from it pretty soon and leave close animations as a progressive enhancement for newer browsers?

I believe we need to be on Safari 17 which will be a while. 
I remember when we tested this there were some issues with the animation. If we can get it to all work on all browsers but maybe Safari <17 doesn't have a close animation, then we should be good.

PR #509, opened 2026-02-11
Add component development process ADR

ianmcburnie:
Formalizes the component development process to prioritize accessibility and reduce drift.

### DIFF
--- a/docs/adr/0002-component-development-foundations.md
+++ b/docs/adr/0002-component-development-foundations.md
@@ -1,4 +1,4 @@
-# 2. Component Development Process
+# 2. Component Development Foundations
 
 **Date:** 2026-02-10
 


agliga (changes requested):
When I look at ADRs to write them or read them, I try to be as unbiased as I can be to convey what decision was made. 
I think the tone of this ADR is too anti-framework. 
I would argue that we shouldn't even be bringing up frameworks at all.

My understanding of the purpose of this ADR is that our components should focus on semantic HTML and CSS first. 
This is all we should really be saying. 
How we get to this shouldn't be the purpose of an ADR. 

Remember, we each work and think differently! The end goal is what is important.

I agree with the ADR and I like a lot of the content. However, some of the "Content" section feels totally out of place.  I think merging this as is can set a bad precedent. Again to reiterate, we need to be as unbiased in our decisions as possible. 

If this seems incorrect, please give me examples of where it wouldn't be!

docs/adr/0002-component-development-process.md
```diff
@@ -0,0 +1,59 @@
...
+To preserve the strengths of our established practice and prevent further drift, we are formalizing the default authoring approach for all new components.
+
+## Decision
+
+We will adopt the following **ground‑up, accessibility‑first component development process** for all new components:
+
+1. **Start with semantic HTML and CSS**
+
+   * Define structure, design hooks, and theming.
+   * Prefer native elements and web standards where appropriate.
```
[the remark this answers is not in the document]
ianmcburnie: I'm trying to keep this as high-level approach as possible, but I can probably work that in. We just have to be sure non-accessible CSS approaches/hacks aren't favoured, so it does open that can of worms

docs/adr/0002-component-development-process.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+For years, the team has followed a de facto ground-up approach when building components: starting with semantic HTML and CSS, validating accessibility and interaction requirements, and then layering in behavior and framework-specific implementations. This practice has contributed to strong accessibility outcomes, framework agnosticism, and long-term maintainability.
+
+Recently, we have observed a tendency to begin directly within frameworks (e.g., Marko, React) for speed or convenience. While efficient in the short term, this shift risks reducing accessibility transparency, introducing framework-specific assumptions too early, and making components harder to reason about outside of a specific implementation.
```
LuLaValva: > Can you explain what the bias is and where it occurs?

@ianmcburnie it's difficult to articulate exactly what I mean, but the way language is laid out here is likely what @agliga  meant by bias.

I understand this is the `context` section so it's supposed to feel more conversational, but the setup here feels antagonistic towards workflows that don't match what you expect. There's a blurry, subjective line between **guidance** and **dogma**, but my gut tells me that this is the latter.

The language here feels like it alienates people who want to use any tools besides those which we have used "for years". From my perspective, the purpose of this ADR is to establish that our components should be written with semantic HTML, CSS, and a11y in mind _before_ behavior is added.

agliga: I think we’re all in agreement that we should start from HTML and CSS. 
I think that’s all the ADR should say.

agliga (approved):

ArtBlue (dismissed):
LGTM

PR #515, opened 2026-02-12
docs(adr): add Percy automation ADR

HenriqueLimas:
## Description

Created ADR following changes from #414

### DIFF
--- a/docs/adr/0003-visual-regression-testing-tool.md
+++ b/docs/adr/0003-visual-regression-testing-tool.md
@@ -8,7 +8,7 @@ Accepted
 
 ## Context
 
-We need visual regression testing to catch unintended UI changes across our component library. With approximately 990 visual snapshots across our components, we needed to evaluate different visual regression testing tools to determine the best fit for our needs.
+We need visual regression testing to catch unintended UI changes across our component library. We have approximately 990 snapshots, and Percy snapshots were previously executed manually. That process was paused because full runs were costly and slow, and running the full suite for every PR was not sustainable.
 
 The primary options considered were:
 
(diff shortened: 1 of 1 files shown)


docs/adr/0003-visual-regression-testing-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+We will use Percy for visual regression testing.
+
+Key factors in this decision:
+
+1. **Existing infrastructure**: We already have an enterprise Percy plan at eBay, eliminating procurement and setup overhead
+2. **Cross-browser testing**: Percy provides testing across multiple browsers (Chrome, Firefox, Safari, Edge), which visual-html does not natively support
+3. **Cost management**: While Percy has costs associated with our 990 snapshots, we can manage this through partial builds that only run snapshots for changed components
+4. **Cloud infrastructure**: Percy's cloud-based service reduces our CI infrastructure burden compared to self-hosting visual-html
+5. **Integration**: Percy integrates well with our GitHub Actions workflow and provides PR integration
```
agliga: I think these three aren't good points imo. 
The partial builds isn't a reason to use percy (we can achieve those anyways)
The reducing burden to our CI isn't even a good point.
The integration with our Github Actions workflow isn't Percy specific. 
I would keep 1 and 2
I would do this:

```suggestion
3. **Accurate**: Percy has been found to be accurate in finding differences between snapshots. Some other tools tend to show a lot of false positive
4. **Cloud infrastructure**: Percy's cloud-based service provides a good UI to quickly see the visual differences and approve the build.
```

docs/adr/0003-visual-regression-testing-tool.md
```diff
@@ -0,0 +1,45 @@
...
+- **visual-html**: An eBay open-source tool for visual regression testing (https://github.com/eBay/visual-html)
+- **Vitest snapshot testing**: Built-in snapshot capabilities in our existing test framework
+
+## Decision
+
+We will use Percy for visual regression testing.
+
+Key factors in this decision:
+
+1. **Existing infrastructure**: We already have an enterprise Percy plan at eBay, eliminating procurement and setup overhead
```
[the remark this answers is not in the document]
agliga: I think any mention of what kind of plan we pay for shouldn't be put out in the open, especially in an ADR.
By saying we have an `existing account` it keeps it ambiguous which I think is good in this case. 
It doesn't take away from anything but rather it doesn't reveal too much.

docs/adr/0003-visual-regression-testing-tool.md
```diff
@@ -0,0 +1,44 @@
...
+2. **Cross-browser testing**: Percy provides testing across multiple browsers (Chrome, Firefox, Safari, Edge), which visual-html does not natively support
+3. **Accurate**: Percy has been found to be accurate in finding differences between snapshots. Some other tools tend to show a lot of false positive
+4. **Cloud infrastructure**: Percy's cloud-based service provides a good UI to quickly see the visual differences and approve the build.
+
+## Consequences
+
+**Positive:**
+
+- Cross-browser visual regression coverage without additional infrastructure
+- Leverages existing eBay Percy enterprise plan
```
HenriqueLimas:
```suggestion
```

agliga (dismissed):

ianmcburnie (approved):

ArtBlue (changes requested):
Looks great. Just a couple of minor things.

PR #525, opened 2026-02-19
feat(evo-react): add EvoButton component

HenriqueLimas:
## Description

- Add new `EvoButton` component following EbayButton component but with new structure
- Make sure types are automatically detected based on `href` presence or not
- Improve storybook documentation (from original EbayButton)
- Automatically import `@ebay/skin/button` file

## Notes

- Some dependencies like ProgressSpinner and ChevronIcon will be implemented later and added to the EvoButton component, currently hard coded
- To support React Router and other framework links, we will also add a `as=` property, that will be in a different PR

## Screenshots

Screenshot 2026-02-18 at 6 12 46 PM

## Checklist

### DIFF
--- a/.changeset/add-evo-button.md
+++ b/.changeset/add-evo-button.md
@@ -1,0 +1,5 @@
+---
+"@evo-web/react": patch
+---
+
+feat(evo-react): add EvoButton component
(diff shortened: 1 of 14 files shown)


.changeset/add-evo-button.md
```diff
@@ -0,0 +1,5 @@
+---
+"@evo-web/react": patch
```
agliga: Not sure if this is a problem, but lets do a 0.x release cycle for this package. 
The reason is because we can clean up and change a bit of the APIs until we do a 1.0 release.

HenriqueLimas: Yeah I wanted to do the first `0.0.1` first and then we will do eventually `0.1.0` until we have all the comments and it is stable.

PR #545, opened 2026-03-02
fix(video): prevent auto-fullscreen on mobile devices

LuLaValva:
## Description

On iOS, clicking "play" on the video automatically moved it to full screen. After this change, fullscreen only happens when the user manually clicks on the fullscreen button.

**Bonus change**: ensure that storybook controls for video are working.


saiponnada (approved):
LGTM!

PR #547, opened 2026-03-03
fix(video): null reference error on play/pause

patrickufer:
TLDR: We shouldn't call `alignSeekbar()` for compact video variant that doesn't render seekbar

- Also sneaks in a fix to use a valid `thumbnail` asset for compact video examples as it seems the previous one expired.

## Description

`ebay-video` compact variant is erroring in alignSeekbar() on [this line](https://github.com/ebay/evo-web/blob/main/packages/ebayui-core/src/components/ebay-video/component.ts#L188) with:
```
video.stories-C02IUq9A.js:22 Uncaught TypeError: Cannot read properties of null (reading 'style')
    at b.alignSeekbar (video.stories-C02IUq9A.js:22:7579)
    at b.handlePlaying (video.stories-C02IUq9A.js:22:7899)
```

`.shaka-range-container` is not on the page with compact variant (no seekbar).

This has unfortunate side-effect that it then never reaches play emit a few lines down.

### DIFF
--- a/packages/ebayui-core/src/components/ebay-video/component.ts
+++ b/packages/ebayui-core/src/components/ebay-video/component.ts
@@ -173,7 +173,7 @@ class Video extends Marko.Component<Input, State> {
     }
 
     alignSeekbar() {
-        if (this.el && this.input.layout !== "compact") {
+        if (this.el) {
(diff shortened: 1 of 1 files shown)


LuLaValva (approved):
Woohoo! Thanks for the quick fix

PR #554, opened 2026-03-06
Alert dialog tests

reidmore-online:
## Description

Created unit and interactions tests for ebay-alert dialog. 

There are 10 unit tests for Marko and React, one test is currently skipped on React due to a JSDOM issue where transition events are not supported, resulting in test failure (any tips on fixing that would be appreciated). All other tests pass.

There are 5 interaction tests for both, currently in react the last test is failing due to an unknown focus loss issue (again, tips appreciated). All other tests pass.

## Notes

For Storybook interactions, we have added @storybook/test to the project, but it is throwing a dependency error every time you start storybook. It does not seem to affect anything actually working, but it is a mystery why it seems dependent on storybook v8 and not v10.


packages/ebayui-core/src/components/ebay-alert-dialog/test/accessibility.browser.js
```diff
@@ -0,0 +1,156 @@
...
+
+beforeAll(() => fastAnimations.start());
+afterAll(() => fastAnimations.stop());
+afterEach(cleanup);
+
+/** @type import("@marko/testing-library").RenderResult */
+let component;
+let user;
+
+describe("accessibility", () => {
```
saiponnada: I know initially we said to have separate files, now that we have better tests I am okay to merge them with current test file.

src/storybook-tests/alert-dialog-interactions.ts
```diff
@@ -0,0 +1,50 @@
...
+        await waitFor(() => {
+            const button = canvas.getByRole("button", { name: "Open Alert Dialog" });
+            expect(button).toHaveFocus();
+        });
+    });
+
+    if (additionalTests) {
+        await additionalTests({ canvasElement, step });
+    }
+};
```
saiponnada: We need to add an extra line

PR #569, opened 2026-03-18
fix(icons): fixed the misalignment of id:viewBox in some icons

### LINKED ISSUE #564: bug: incorrect icon viewports
It looks like there is a viewport mismatch between icon IDs and viewports here: https://github.com/eBay/evo-web/blob/main/packages/skin/src/svg/icons.svg?short_path=e0bf708#L894-L907

Looking at the details, it looks like the individual `SVG` files we get from design also had the incorrect viewports, so what likely happened is that these errors trickled into the symbols from the automated icon creation script.

For this issue - 

1. At minimum, correct the individual icon and symbol viewports manually.
2. Add a layer of checks in our script to flag these so they don't make it through a commit. Colored icons need to be excluded from that mechanism since their can potentially differ.

ArtBlue:
- Fixes #564 

# Description
Fixed icons with incorrect viewBox attributes in their source SVG files. 

✅ icon-arrows-expand-16.svg - Changed from 24x24 to 16x16
✅ icon-arrows-expand-24.svg - Changed from 16x16 to 24x24

✅ icon-feedback-error-16.svg - Changed from 24x24 to 16x16
✅ icon-feedback-error-24.svg - Changed from 16x16 to 24x24

## Checklist

### DIFF
--- a/.changeset/violet-spoons-sing.md
+++ b/.changeset/violet-spoons-sing.md
@@ -1,0 +1,5 @@
+---
+"@ebay/skin": patch
+---
+
+fix(icons): fixed the misalignment of id:viewBox in some icons


saiponnada (dismissed):
Overall looks good to me. 

Only question is why `icon-feedback-error-16.svg` does not have padding(close to the borders) where as all other icons do have some padding?

PR #580, opened 2026-03-25
fix(filter-chip): replace aria-pressed with CSS class for menu button selected state

### LINKED ISSUE #452: filter-chip: incorrectly uses aria-pressed for button elements
## Problem

The filter chip component uses `aria-pressed` to communicate selection state for `<button>` elements, which is semantically incorrect.

## Solution

Replace `aria-pressed` with visually hidden `.clipped` text for button filter chips, matching the anchor implementation.

**Before:**
```html
<button class="filter-chip" aria-pressed="true">
  <span class="filter-chip__text">Football</span>
</button>
```

**After:**
```html
<button class="filter-chip">
  <span class="filter-chip__text">
    Football
    <span class="clipped"> - Filter Applied</span>
  </span>
</button>
```

We still need to iron our the appropriate text inside the clipped span.

### Screenshots

_No response_

### Figma link

_No response_

- [ ] Supports dark mode.

voidmatcha:
- Fixes #452

## Notes

Menu filter-chips used `aria-pressed` to show selected state. That's the wrong attribute here - `aria-pressed` is for buttons that toggle their own state on/off, but a menu chip reflects a selection made inside another UI element. Screen readers were announcing a pressed state that had nothing to do with the button itself.

Two changes:

1. CSS — added `filter-chip--selected` class as a selector for selected state and animation, alongside the existing `aria-pressed="true"`. Menu chips can now drop `aria-pressed` without losing visual feedback.
2. Stories — removed `aria-pressed` from all menu button chip examples, added `filter-chip--selected` class and `<span class="clipped">filter applied</span>` so screen readers announce filter state correctly.

Non-menu chips (no `aria-expanded`) still use `aria-pressed` as before.

**As-Is**
`<button class="filter-chip" aria-pressed="true">`
**To-Be**
`<button class="filter-chip filter-chip--selected">` + `<span class="clipped">filter applied</span>`

## Checklist

### DIFF
--- a/.changeset/loose-lizards-drop.md
+++ b/.changeset/loose-lizards-drop.md
@@ -1,5 +1,8 @@
 ---
 "@ebay/skin": patch
+"@ebay/ebayui-core": patch
+"@ebay/ebayui-core-react": patch
+"@evo-web/marko": patch
 ---
 
 replace aria-pressed with CSS class for menu button selected state
--- a/.changeset/loose-lizards-drop.md
+++ b/.changeset/loose-lizards-drop.md
@@ -1,0 +1,5 @@
+---
+"@ebay/skin": patch
+---
+
+replace aria-pressed with CSS class for menu button selected state
--- a/packages/ebayui-core-react/src/ebay-filter-chip/filter-chip.tsx
+++ b/packages/ebayui-core-react/src/ebay-filter-chip/filter-chip.tsx
@@ -81,11 +81,13 @@ const EbayFilterChip: FC<EbayFilterChipProps> = ({
         }
     };
 
+    const isMenu = variant === "menu";
+
     const classNames = classnames(
         "filter-chip",
         {
             "filter-chip--expressive": variant === "expressive",
-            "filter-chip--selected": isAnchor && selected,
+            "filter-chip--selected": (isAnchor || isMenu) && selected,
         },
         className,
     );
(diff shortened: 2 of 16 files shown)


saiponnada (commented):
Thanks @dididy for taking this up. Few things pending,

- We need examples in skin stories for expressive variant(with `aria-pressed` but no `aria-expanded`)
- Also update the docs under `filter-chip/css+page.marko`
- Implementations on both Marko and React and add corresponding tests.

If you want to close this out, we’re happy to take it on—especially if working across the frameworks isn’t convenient on your side.

voidmatcha:
Addressed all the review feedback!
- Removed `aria-pressed` from menu variant in Marko/React, using `filter-chip--selected` class + clipped text instead
- Updated docs, stories, and dist
- Rebased on latest main to pick up recent merges
- Set up GPG signing for verified commits as suggested https://github.com/eBay/evo-web/pull/577#issuecomment-4215954005

The build has some `ebay-date-textbox` snapshot failures, but the same thing happens on main — not related to this PR.

.changeset/loose-lizards-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/skin": patch
```
saiponnada, line 2: We also need to include following packages

@ebay/ebayui-core 
@ebay/ebayui-core-react 
@evo-web/marko

src/routes/_index/components/filter-chip/css+page.marko
```diff
@@ -315,6 +318,9 @@
       <svg class="filter-chip__trailing icon icon--12" width="12" height="12">
         <use href="#icon-chevron-down-16"/>
       </svg>
+      <span class="clipped">
+        filter applied
```
saiponnada: Lets be consistent across using title case "Filter Applied"

packages/ebayui-core-react/src/ebay-filter-chip/filter-chip.tsx
```diff
@@ -81,11 +81,13 @@ const EbayFilterChip: FC<EbayFilterChipProps> = ({
         }
     };
 
+    const isMenu = variant === "menu";
```
saiponnada: `isMenu` can be moved up and can be simplified further. We can also leverage isMenu in handleClick

```
    const isMenu = variant === "menu";
    const isAnchor = !!href && !isMenu; 
```

saiponnada (changes requested):
Overall LGTM! Thanks for making these changes. We have minor knits to fix and good to merge IMO.

voidmatcha:
All addressed. thanks for the review! 🙏

saiponnada (approved):
LGTM!

PR #601, opened 2026-04-02
refactor(skin): gitignore build files

LuLaValva:
Ignore all files in `dist` and use a postinstall hook to build instead

This means initial setup takes a bit longer, but diff view will be much easier to manage and changes will be more minimal

Not sure if this is a good idea so keeping as a draft for now


ArtBlue (commented):
If there are no downstream negative implications of this change, it seems like a no-brainer. Overall, it looks good. Just a couple of concerns/questions.

ArtBlue (changes requested):
@LuLaValva , since Percy builds require diffs in `dist`, ignoring the path will cause no Percy builds to be fired. If this is the only material change in this PR, let's just close it as won't do.

LuLaValva:
Okay, closing for now since it's out of scope to also change all of the Percy pipelines. We'll come back to this later with a bigger plan about what parts we are safe to prune or move to build/CI time

PR #622, opened 2026-04-14
feat(evo-marko): alert & confirm dialog

LuLaValva:
## Description

- Add `<evo-alert-dialog>` and `<evo-confirm-dialog>`

## Notes

- The real question here is should these be separate components? Since they are so common and there is slightly different behavior for each of them I think there's value in keeping them apart, but this is up for debate.
- I also modified some AI instructions to make sure code follows best practices moving forward

## Screenshots

image
image

### DIFF
--- a/.changeset/funky-pears-occur.md
+++ b/.changeset/funky-pears-occur.md
@@ -1,0 +1,5 @@
+---
+"@evo-web/marko": patch
+---
+
+Confirm & Alert Dialog
--- a/.claude/skills/evo-migrate-marko/SKILL.md
+++ b/.claude/skills/evo-migrate-marko/SKILL.md
@@ -132,6 +132,7 @@ Key changes:
 - **Remove `Omit<..., 'on${string}'>`** -- Marko 6 handles events natively via spread.
 - **Remove custom `on-*` event props** from the interface -- consumers bind native DOM events directly or use two-way binding callbacks (e.g., `valueChange`, `openChange`, `indexChange`).
 - **Use camelCase** for all prop names (not kebab-case).
+- **Use `Marko.HTML.*`** for element types -- e.g., `Marko.HTML.H2`, `Marko.HTML.Div`, `Marko.HTML.Button`. Do **not** use `Marko.Input<"h2">` or `Marko.Input<"div">`.
 - **Omit only props the component hardcodes** (e.g., `Omit<Marko.HTML.Input, "type" | "role">`).
 
 ### HTML attribute pass-through
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,6 +86,10 @@ HTML Semantic Structure → @ebay/skin (CSS/BEM) → Framework Components → In
 - ✅ Attribute values containing `>` MUST be wrapped in parentheses: `<const/x=(a > b ? 1 : 0)>`
 - ❌ Never: `<const/x=a > b ? 1 : 0>` (the `>` is parsed as the tag close)
 
+**Marko 6 Tag Variable Locality:**
+
+Declare `<const/>`, `<let/>`, `<id/>`, and other tag variables close to where they are first used, not grouped at the top. Exception is variables needed in multiple distant locations.
+
 **React Package Differences:**
 
 - `ebayui-core-react`: Requires `React.forwardRef` wrapper
--- a/packages/evo-marko/src/tags/evo-alert-dialog/index.marko
+++ b/packages/evo-marko/src/tags/evo-alert-dialog/index.marko
@@ -1,0 +1,79 @@
+import { type Input as ButtonInput } from "<evo-button>";
+
+export interface Input extends Omit<Marko.HTML.Dialog, "open" | "closedby" | "role" | "aria-labelledby"> {
+  open?: boolean;
+  openChange?: (o: boolean) => void;
+  header: Marko.AttrTag<Marko.HTML.H2 & { as?: string }>;
+  confirm: Marko.AttrTag<ButtonInput>;
+}
+
+<const/{
+  open: inputOpen,
+  openChange,
+  class: inputClass,
+  header,
+  confirm,
+  content,
+  onCancel,
+  onAnimationEnd,
+  ...htmlInput
+}=input>
+
+<let/open:=input.open>
+<id/headerId=header.id>
+
+<script>
+  if (open && !$dialog().open) {
+    $dialog().showModal();
+  }
+</script>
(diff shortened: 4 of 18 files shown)


packages/evo-marko/src/tags/evo-confirm-dialog/index.marko
```diff
@@ -0,0 +1,92 @@
...
+  <div id=mainId class="dialog__main">
+    <${content}/>
+  </div>
+  <div class="dialog__footer">
+    <const/{ onClick: rejectOnClick, ...rejectInput }=reject>
+    <evo-button
+      ...rejectInput
+      onClick(e, el) {
+        open = false;
+        rejectOnClick && rejectOnClick(e, el);
```
saiponnada, line 77: One thing to iron out is, in legacy implementation we always handled [ESC] - rejection & Cancel as same.  Here are we separating those two? For simplicity, we might need to merge those two; lets use rejection during cancel as well.

packages/evo-marko/src/tags/evo-confirm-dialog/index.marko
```diff
@@ -0,0 +1,92 @@
...
+<script>
+  if (open && !$dialog().open) {
+    $dialog().showModal();
+  }
+</script>
+
+<dialog/$dialog
+  ...htmlInput
+  open=null // tell Marko it's not a part of the spread because the browser does DOM manipulation
+  role="alertdialog"
```
saiponnada, line 37: I have a philosophical question if we should still use `role="alertdialog"`for confirm too. I get it makes sense for alert dialog and we have been using this way till now. Let me get back to you on that.

packages/evo-marko/src/tags/evo-confirm-dialog/confirm-dialog.stories.ts
```diff
@@ -0,0 +1,68 @@
+import { buildExtensionTemplate } from "../../common/storybook/utils";
+import { type Meta } from "@storybook/marko";
+import Readme from "./README.md";
+import Component, { type Input } from "./index.marko";
+import DefaultTemplate from "./examples/default.marko";
+import DefaultCode from "./examples/default.marko?raw";
+
+export default {
```
saiponnada, line 8: Any reason why we are no longer logging events in Action tab today? Is it WIP?

saiponnada (commented):
LGTM! minor things to iron out.

packages/evo-marko/src/tags/evo-confirm-dialog/index.marko
saiponnada: > An alert dialog is a [modal dialog](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) that interrupts the user's workflow to communicate an important message and acquire a response.

So it make sense to continue using `alertdialog` I guess. Ignore this then.

packages/evo-marko/src/tags/evo-confirm-dialog/confirm-dialog.stories.ts
LuLaValva: We aren't logging any events in the action tab yet, but maybe we should. I'll look into logging events to all of the components in a follow-up.

One note: all of the events for this component (so far) are _native HTML events_. We definitely shouldn't log all of those or else we'll see a million `mousemove` and other arbitrary events, but maybe I'll add some of the more important ones to stories.

packages/evo-marko/src/tags/evo-confirm-dialog/index.marko
LuLaValva: Here's the really cool thing about using native dialog—

I changed `open = false` to `$dialog().requestClose()` and now `onCancel` works for both of them!

.changeset/funky-pears-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@evo-web/marko": patch
```
ArtBlue, line 2: This should be a minor

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue, line 1: This is great! Did you use `skill-creator` to create this?

CLAUDE.md
```diff
@@ -86,6 +86,10 @@ HTML Semantic Structure → @ebay/skin (CSS/BEM) → Framework Components → In
 - ✅ Attribute values containing `>` MUST be wrapped in parentheses: `<const/x=(a > b ? 1 : 0)>`
 - ❌ Never: `<const/x=a > b ? 1 : 0>` (the `>` is parsed as the tag close)
 
+**Marko 6 Tag Variable Locality:**
```
ArtBlue, line 89: Nice! ❤️

ArtBlue (commented):
Looking good to me! Just version type.

.changeset/funky-pears-occur.md
LuLaValva: No, we are still on `0.x.x` for `@evo-web/marko` so we should use patch for non-breaking changes and minor for breaking.

.claude/skills/evo-migrate-marko/SKILL.md
LuLaValva: I believe so, I made it for the last migration. There is still a lot of back-and-forth for each component, but I hope it will improve as I modify each time

packages/evo-marko/src/tags/evo-alert-dialog/index.marko
```diff
@@ -0,0 +1,75 @@
...
+  aria-modal="true"
+  closedby="none"
+  aria-labelledby=headerId
+  class=[
+    "dialog",
+    "dialog--narrow",
+    !open && "dialog--close",
+    inputClass,
+  ]
+  onAnimationEnd(e, el) {
```
saiponnada, line 43: Does team have ability to pass their custom animations? If yes, should we add reduced motion to respect user animations. We should discuss this further and add support if needed in following PR.

saiponnada (approved):
LGTM!

packages/evo-marko/src/tags/evo-alert-dialog/index.marko
LuLaValva: Custom animations are valid, but if there is ever _no_ animation then we will run into problems. My recommendation to people who have a jarring animation is to make a minimal one under `prefers-reduced-motion` but not remove it entirely, but in the future we can talk about adding support for no animations

PR #629, opened 2026-04-17
Audited and Improved Evo Migrate Marko Skill

ArtBlue:
## Description

I put the current `evo-migrate-marko` skill through an audit using the `skill-creator` skill and used the conservative set of changes it suggested to slightly improve the skill. 

## Notes

There was a broader set of suggested changes to the skill that did not dramatically improve the end result of the skill. However, it did reduce the token usage and took less time to run. The broader audit also included evals for baselines on performance and efficiency of the skill. I left that out of this PR, too. If desired, that can be done as a separate PR by running `skill-creator` audit on the skill at any time.

This is the benchmark dashboard comparing the efficiency of the current skill with the potential efficiency gains from the revised skill (with all possible improvements).

<kbd>image</kbd>

### DIFF
--- a/.claude/skills/evo-migrate-marko/SKILL.md
+++ b/.claude/skills/evo-migrate-marko/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: evo-migrate-marko
-description: Migrate a component from @ebay/ebayui-core (Marko 5) to @evo-web/marko (Marko 6). Receives the ebayui-core component name as the argument (e.g. /evo-migrate-marko ebay-button).
+description: Migrate a component from @ebay/ebayui-core (Marko 5) to @evo-web/marko (Marko 6). Receives the ebayui-core component name as the argument (e.g. /evo-migrate-marko ebay-button). Use this skill whenever the user mentions migrating, porting, converting, or moving any ebay-* Marko component to evo-marko, even if they don't explicitly invoke the skill by name.
 ---
 
 # Migrate ebayui-core (Marko 5) to evo-marko (Marko 6)
@@ -27,10 +27,12 @@ This is an ACTIVE PROCESS, before making any decision you should consult with th
    - Complex (floating UI positioning): `packages/evo-marko/src/tags/evo-tooltip/index.marko`
 3. If the Marko 5 component uses shared base classes or utilities from `src/common/`, read those too.
 4. If the Marko 5 component uses `makeup-*` libraries, check whether an evo-marko headless replacement exists:
-   - `makeup-roving-tabindex` / `makeup-active-descendant` -> `<evo-roving-tabindex>` (in `src/tags/tags/`)
+   - `makeup-roving-tabindex` -> `<evo-roving-tabindex>` (in `src/tags/tags/`) — use when focus physically moves into list items (tabs, menus, listboxes where options are focusable)
+   - `makeup-active-descendant` -> **do NOT use `<evo-roving-tabindex>`** for combobox-like patterns where focus stays on the input. Instead, manage `aria-activedescendant` directly: `<let/activeIndex=-1>` on the input with `aria-activedescendant=(activeIndex >= 0 ? \`${listId}-${activeIndex}\` : undefined)`.
    - `makeup-expander` + `@floating-ui/dom` -> `<evo-expander>` (in `src/tags/tags/`)
    - `makeup-floating-label` -> `<evo-floating-label>` (in `src/tags/tags/`)
    - `makeup-typeahead` -> `<evo-typeahead>` (in `src/tags/tags/`)
+   - `makeup-focusables` -> **no evo replacement**; use native `querySelectorAll` with a standard focusable selector instead
 
 ---
 
@@ -66,6 +68,16 @@ packages/evo-marko/src/tags/evo-{name}/
 ```
 
 **Key difference from ebayui-core:** Everything lives in `index.marko`. There is no separate `component.ts`, `component-browser.ts`, or `marko-tag.json`.
+
+### README.md content
+
+Write a `README.md` that covers:
+
+- **One-line description** of what the component does
+- **Basic usage example** (copy-pasteable Marko snippet)
+- **Props table** — name, type, default, description (derive from the `Input` interface)
+- **Accessibility notes** — any keyboard interactions, ARIA roles/attributes, screen reader behavior
+- **Breaking changes from ebayui-core** — renamed props, removed props, changed event signatures
 
 ---
 


ArtBlue:
Ignoring all Copilot suggestions from this PR as it doesn't have enough Claude context and is approaching things from a human readability angle rather than skill efficiency.

.claude/skills/evo-migrate-marko/SKILL.md
```diff
@@ -27,10 +27,12 @@ This is an ACTIVE PROCESS, before making any decision you should consult with th
    - Complex (floating UI positioning): `packages/evo-marko/src/tags/evo-tooltip/index.marko`
 3. If the Marko 5 component uses shared base classes or utilities from `src/common/`, read those too.
 4. If the Marko 5 component uses `makeup-*` libraries, check whether an evo-marko headless replacement exists:
-   - `makeup-roving-tabindex` / `makeup-active-descendant` -> `<evo-roving-tabindex>` (in `src/tags/tags/`)
+   - `makeup-roving-tabindex` -> `<evo-roving-tabindex>` (in `src/tags/tags/`) — use when focus physically moves into list items (tabs, menus, listboxes where options are focusable)
+   - `makeup-active-descendant` -> **do NOT use `<evo-roving-tabindex>`** for combobox-like patterns where focus stays on the input. Instead, manage `aria-activedescendant` directly: `<let/activeIndex=-1>` on the input with `aria-activedescendant=(activeIndex >= 0 ? \`${listId}-${activeIndex}\` : undefined)`.
```
LuLaValva, line 31: I think this is a hallucination, I haven't migrated combobox yet. It _is_ true that `makeup-active-descendant` _should not_ be replaced by `<evo-roving-tabindex>` though

.claude/skills/evo-migrate-marko/SKILL.md
```diff
@@ -27,10 +27,12 @@ This is an ACTIVE PROCESS, before making any decision you should consult with th
...
    - `makeup-expander` + `@floating-ui/dom` -> `<evo-expander>` (in `src/tags/tags/`)
    - `makeup-floating-label` -> `<evo-floating-label>` (in `src/tags/tags/`)
    - `makeup-typeahead` -> `<evo-typeahead>` (in `src/tags/tags/`)
+   - `makeup-focusables` -> **no evo replacement**; use native `querySelectorAll` with a standard focusable selector instead
```
LuLaValva, line 35: Not necessarily true, we may end up making a Marko equivalent of this

.claude/skills/evo-migrate-marko/SKILL.md
```diff
@@ -67,6 +69,16 @@ packages/evo-marko/src/tags/evo-{name}/
...
 
+### README.md content
+
+Write a `README.md` that covers:
+
+- **One-line description** of what the component does
+- **Basic usage example** (copy-pasteable Marko snippet)
+- **Props table** — name, type, default, description (derive from the `Input` interface)
+- **Accessibility notes** — any keyboard interactions, ARIA roles/attributes, screen reader behavior
+- **Breaking changes from ebayui-core** — renamed props, removed props, changed event signatures
```
LuLaValva, line 80: We shouldn't include breaking changes from eBayUI in the README. We'll figure out a better migration path once I've moved all components over

.claude/skills/evo-migrate-marko/SKILL.md
```diff
@@ -67,6 +69,16 @@ packages/evo-marko/src/tags/evo-{name}/
 
 **Key difference from ebayui-core:** Everything lives in `index.marko`. There is no separate `component.ts`, `component-browser.ts`, or `marko-tag.json`.
 
+### README.md content
+
+Write a `README.md` that covers:
+
+- **One-line description** of what the component does
+- **Basic usage example** (copy-pasteable Marko snippet)
+- **Props table** — name, type, default, description (derive from the `Input` interface)
```
LuLaValva, line 78: We should **NOT** have a props table in the README, it will drift out of sync very quickly. There is no reason to add another maintenance point

LuLaValva (commented):
@ArtBlue I'm not sure I understand how this saved tokens? All of the changes are additions to the skill, not removals

ArtBlue:
> @ArtBlue I'm not sure I understand how this saved tokens? All of the changes are additions to the skill, not removals

The size of the skill is irrelevant. We're not measuring the tokens *in* the skill. The performance evals are based on the efficiency and effectiveness of the `evo-marko-migration` skill, not the length of the skill itself. The evals were done with the existing skill and the skill with the current modifications. The improvements are based on the comparison of those results.

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: @LuLaValva , again, these are the instructions for the skill. Combobox doesn't have to have been migrated yet. This is context into the skill to improve the effectiveness of the skill.

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: @LuLaValva , the future is not very relevant with respect to the skill. It's what is true **right now**. Since this is an ephemeral skill anyway, the future may be wholesale irrelevant. As the codebase exists now, is the statement true for the skill to get enough context to do a better job. This is the only question we should be asking.

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: This is not the README, it's the SKILL. If the statements are true, we should keep them in since their purpose is to help the skill do a better job of migrating components. If they need adjustment because of some nuance, we should adjust them. If they are wholesale inaccurate, then we should remove/exclude them.

.claude/skills/evo-migrate-marko/SKILL.md
LuLaValva: I understand that, but this isn't necessarily the solution we want to use for migrating `makeup-focusables` and I think it is a hallucination

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: This is the SKILL, not the README. I think this is meant to provide a template/pattern. Excluding them will leave it up the model's understanding and will make this non-deterministic.

.claude/skills/evo-migrate-marko/SKILL.md
LuLaValva: This is a part of the skill that is instructing AI how to migrate a README from `ebayui-core` to `evo-marko`. The statements are not true, as these changes should _not_ be made to the README

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: I'm not sure whether it's a hallucination or not, but it is trying to make explicit what is currently implied. What's merely implied creates vagueness and leads to non-deterministic outcomes. That said, your concern is fair. What is the right solution? Maybe we can capture that explicitly.

.claude/skills/evo-migrate-marko/SKILL.md
ArtBlue: Gotcha. Ok. Can you suggest alternative wording of what should go in the README?

LuLaValva: I think we should either remove this section or tell it to copy the README directly and find & replace `ebay-*` to `evo-*`

.claude/skills/evo-migrate-marko/SKILL.md
LuLaValva: We don't know what the right solution is yet because we haven't migrated the combobox component yet. I will add the correct solution when I migrate a component that uses this pattern and have had time to figure out what we should be doing in tags API

ArtBlue:
Closing this so Luke can make the changes himself.

PR #630, opened 2026-04-17
feat(evo-marko): evo-chip

LuLaValva:
- Make `<evo-chip>` component
- This one is _very_ straightforward as it's a display component, should be a quick review


saiponnada (approved):
LGTM!

PR #631, opened 2026-04-17
feat(evo-marko): evo-toast-dialog

LuLaValva:
- Create `<evo-toast-dialog>` component

## Differences from `<ebay-toast-dialog>`

- Use `<dialog>` instead of `<aside>`
  - This means the top-level component is also in the focus order (it wasn't before)
- Use `@animation` instead of `transition`
- `<@close>` attr tag instead of `a11yCloseText`
- controllable `open` (with `openChange`)


saiponnada (approved):
LGTM!

PR #633, opened 2026-04-21
feat(evo-marko): evo-star-rating

LuLaValva:
- Add `<evo-star-rating>` component

## Differences from `<ebay-star-rating>`

- `a11yText` is required with TS
- Receives an integer value (`2.5`) instead of our special data attribute value (`2-5`)
  - This is more future proof, in case we ever support something like fractional stars

### DIFF
--- a/packages/evo-marko/src/tags/evo-star-rating/index.marko
+++ b/packages/evo-marko/src/tags/evo-star-rating/index.marko
@@ -11,12 +11,12 @@ export interface Input extends Omit<Marko.HTML.Div, "aria-label" | "role"> {
     /**
      * Accessible label for the star rating.
      *
-     * English default to be overridden is `"${value} out of 5 stars"`.
+     * English default to be overridden is `"Rating: ${value} out of 5"`.
      */
     a11yText: string;
 }
 
-<const/{ class: inputClass, value = 0, a11yText = `${value} of 5 stars`, ...htmlInput }=input>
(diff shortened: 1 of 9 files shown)


packages/evo-marko/src/tags/evo-star-rating/index.marko
```diff
@@ -0,0 +1,30 @@
...
+    value?: number;
+    /**
+     * Accessible label for the star rating.
+     *
+     * English default to be overridden is `"${value} out of 5 stars"`.
+     */
+    a11yText: string;
+}
+
+<const/{ class: inputClass, value = 0, a11yText = `${value} of 5 stars`, ...htmlInput }=input>
```
saiponnada: Minor knit, lets match the JSDoc & SB default text to be consistent with docs. Our docs say `Rating: 2.5 out of 5`. Worth updating across to be consistent.

```suggestion
<const/{ class: inputClass, value = 0, a11yText = `${value} out of 5 stars`, ...htmlInput }=input>
```

saiponnada (commented):
LGTM, minor knit.

packages/evo-marko/src/tags/evo-star-rating/index.marko
LuLaValva: Thanks! Updated

saiponnada (approved):
LGTM!

PR #635, opened 2026-04-22
feat(evo-testing): create skill for test creation

reidmore-online:
## Description

Based on the previous work for the testing prompt, I turned it into a skill to build tests for evo. It currently is only setup and pointed at ebayui-core/ebayui-core-react, it will need to be updated once the other packages are ready to be included. 

## Notes

It is currently built to produce both unit and interactions tests for components, though right now the interactions tests cause issues, so when they are generated we aren't merging them. I am happy to keep it as is, or edit it to only do unit tests right now and develop a separate skill for the interactions tests when those are ready. 

Things I'd love feedback on: 
- Instead of linking out to the evo-web site for docs, is it possibly better to link into the docs, since they are also in the repo? 
- Examples are currently part of the skill file, but I also considered making them reference files to be called as needed, which approach makes more sense?

### DIFF
--- a/.claude/skills/evo-testing/SKILL.md
+++ b/.claude/skills/evo-testing/SKILL.md
@@ -143,7 +143,7 @@ Ask questions about any tests or requirements that there is low confidence about
 
 1. **Marko a11y-focused browser tests:** add accessibility tests to `test.browser.js` and group according to test focus.
 2. **React a11y-focused tests:** add accessibility tests to `*.spec.tsx` and group according to test focus.
-3. **Storybook plays:** Marko → `ebay-{component}.stories.ts` (or the component’s `*.stories.ts`). React → `__tests__/index.stories.tsx` (or colocated stories).
+3. **Storybook plays:** Marko → `src/components/ebay-{component}/{component}.stories.ts` (for example, `src/components/ebay-button/button.stories.ts`). React → `__tests__/index.stories.tsx` (or colocated stories).
 4. Organize by Click / Keyboard / Focus / ARIA.
 5. Use **`beforeEach`** for nested setup (given/when/then).
 6. Prefer **`userEvent`** from `@testing-library/user-event` (or `vitest/browser` where the existing file already does) over ad-hoc `pressKey` helpers.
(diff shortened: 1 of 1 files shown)


LuLaValva (approved):
Thanks for adding this @reidmore-online! Since the only change is a skill, I'm comfortable merging as-is and we can tweak the specific details when there are bugs in tests created using it.

ArtBlue (changes requested):
Can we please audit this skill using `skill-creator` to see if there are opportunities for improvement?

reidmore-online:
> Can we please audit this skill using `skill-creator` to see if there are opportunities for improvement?

Sorry, I should have mentioned I did this step :). I used skill-creator to help me take the old `testing.md` file from previous experiments and turn it into the skill, I then asked it to evaluate the skill and that's the current state. The only thing I did not have it do was test, because I don't really have a comparison to test it with and I think the output would be hard to quantitatively assess.

ArtBlue:
> > Can we please audit this skill using `skill-creator` to see if there are opportunities for improvement?
> 
> Sorry, I should have mentioned I did this step :). I used skill-creator to help me take the old `testing.md` file from previous experiments and turn it into the skill, I then asked it to evaluate the skill and that's the current state. The only thing I did not have it do was test, because I don't really have a comparison to test it with and I think the output would be hard to quantitatively assess.

@reidmore-online , awesome! Did you want to check the evals in to establish a baseline for the skill performance or do you think it's too early for that (if this is still beta)?

reidmore-online:
> > > Can we please audit this skill using `skill-creator` to see if there are opportunities for improvement?
> > 
> > 
> > Sorry, I should have mentioned I did this step :). I used skill-creator to help me take the old `testing.md` file from previous experiments and turn it into the skill, I then asked it to evaluate the skill and that's the current state. The only thing I did not have it do was test, because I don't really have a comparison to test it with and I think the output would be hard to quantitatively assess.
> 
> @reidmore-online , awesome! Did you want to check the evals in to establish a baseline for the skill performance or do you think it's too early for that (if this is still beta)?

I think it's still too early, in my other experiments with evals they seem to be most helpful when you have something to compare to, and I just don't really have that yet. I am curious if I make variants of these skill that reflect the areas I have questions about, then I can possibly run evals between the different versions to see what ends up performing best.

ArtBlue (approved):

ArtBlue:
Evals are also helpful once you start making adjustments before committing potential improvements. In some cases, they are absolutely critical right off the bat. Probably not here.

PR #639, opened 2026-04-22
feat(evo-marko): Break apart `evo-textbox`

LuLaValva:
Break `<evo-textbox>` into `<evo-input>` and `<evo-textarea>`, and did some cleanup on each component


HenriqueLimas:
I think the only one that should change is `textbox` to `input`. 
The others should stay the same `NumberInput`, `FilterInput` aligns more with all other name convention for the other components like Lightbox**Dialog**, Menu**Button** etc

LuLaValva:
Yeah I think you're right @HenriqueLimas, I removed those changes. PR has been repurposed just to breaking apart `<evo-textbox>`

packages/evo-marko/src/tags/evo-textarea/style.ts
```diff
@@ -0,0 +1,2 @@
+import "@ebay/skin/floating-label";
```
HenriqueLimas: Maybe we could move floating-label import to the evo-floating-label instead.

packages/evo-marko/src/tags/evo-input/index.marko
```diff
@@ -41,73 +36,67 @@ export interface Input extends Marko.HTML.Input {
...
                 focused = true;
-                onFocus && onFocus(e, target);
+                (onFocus || null)?.(e, target);
             }
-            id=textboxId
+            id=inputId
             class="textbox__control"
-            type=!multiline && (type || "text")
+            placeholder=showPlaceholder && placeholder
+            type=type
```
HenriqueLimas: We probably don't need this anymore as it will be part of `htmlInput` now

packages/evo-marko/src/tags/evo-textarea/index.marko
```diff
@@ -0,0 +1,74 @@
...
+                "textbox--readonly" /** end remove after `:has` support */,
+            inputSize === "large" && "textbox--large",
+            fluid && "textbox--fluid",
+            inputClass,
+        ]>
+        <textarea
+            ...htmlInput
+            onBlur(e, target) {
+                focused = false;
+                (onBlur || null)?.(e, target);
```
HenriqueLimas, line 60: if we are using optional chaining, do we need the `|| null`? Do we accept `onBlur=false` in the types?

packages/evo-marko/src/tags/tags/evo-floating-label/index.marko
```diff
@@ -1,6 +1,7 @@
 export interface Input {
-    id?: Marko.HTMLAttributes["id"]
+    forId?: Marko.HTMLAttributes["id"]
```
HenriqueLimas: I think we should use `htmlFor` as this is the spec name for labels too https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor

packages/evo-marko/src/tags/tags/evo-floating-label/index.marko
```diff
@@ -38,16 +39,16 @@ export interface Input {
...
-                floatingInline && "floating-label__label--inline",
-                !floatingInline && "floating-label__label--focus",
+                !isFloating && "floating-label__label--inline",
+                isFloating && "floating-label__label--focus",
             ]>
             ${floatingLabel}
         </label>
     </if>
-    <${content}/>
+    <${content} showPlaceholder=!floatingLabel || isFloating/>
```
HenriqueLimas, line 53: THIS IS 🤯 !

packages/evo-marko/src/tags/evo-textarea/index.marko
LuLaValva: Yeah, TS expects to also receive `false`. I've been using `onBlur && onBlur(e, target)` instead in other components, switched to this.

packages/evo-marko/src/tags/evo-input/index.marko
LuLaValva: Great catch

packages/evo-marko/src/tags/evo-textarea/style.ts
LuLaValva: Great point, I moved it.

packages/evo-marko/src/tags/evo-textarea/index.marko
```diff
@@ -0,0 +1,74 @@
+export interface Input extends Marko.HTML.TextArea {
+    inputSize?: "regular" | "large";
```
HenriqueLimas, line 2: I think we don't need this on textarea, we don't have a specific height for large/regular

HenriqueLimas (approved):

PR #640, opened 2026-04-23
feat(evo-react): add evo-details component

HenriqueLimas:
- Fixes #

## Description

Migrates `ebay-details` from `@ebay/ebayui-core-react` to `@evo-web/react` as a new compound component, and adds the missing `leading` attribute tag to `evo-details` in `@evo-web/marko`.

### `@evo-web/react` — new `evo-details` component

Uses a compound sub-component API backed by React context instead of flat props or `React.Children` scanning (see ADR 0005):

```tsx
<EvoDetails size="small" alignment="center" onToggle={handler}>
  <EvoDetailsSummary>
    <EvoDetailsLeading><EvoIconLightbulb16 /></EvoDetailsLeading>
    <EvoDetailsLabel>Show me the details!</EvoDetailsLabel>
  </EvoDetailsSummary>
  <EvoDetailsContent>Content</EvoDetailsContent>
</EvoDetails>
```

Exported components: `EvoDetails`, `EvoDetailsSummary`, `EvoDetailsLeading`, `EvoDetailsLabel`, `EvoDetailsContent`.

### `@evo-web/marko` — `leading` attribute tag on `evo-details`

Adds the `leading` attribute tag that was present in `@ebay/ebayui-core` (Marko 5) but was missed during the Marko 6 migration.

## Notes

- ADR 0005 documents the compound component composition pattern and why `React.Children` / `React.cloneElement` were ruled out.
- `evo-migrate-react` and `evo-app-migrate-react` skills updated with migration guidance for this component.
- The evo-react API intentionally diverges from evo-marko's `<@summary>` attribute tag pattern, which has no direct React equivalent.

## Checklist


packages/evo-react/src/evo-details/details.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: EvoDetailsProps) {
+  const handleToggle =
+    onToggle &&
+    ((event: SyntheticEvent<HTMLDetailsElement>) => {
+      onToggle(event, { open: event.currentTarget.open });
+    });
+
+  return (
+    <DetailsContext value={{ size, alignment }}>
+      <details
```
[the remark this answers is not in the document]
HenriqueLimas: Not on React 19 my friend! This is valid.

.claude/skills/evo-app-migrate-react/SKILL.md
```diff
@@ -55,6 +55,60 @@ If a component is not listed below, it has **not been migrated yet** — keep us
 
 No prop changes. Global renames (Step 2) are sufficient.
 
+### `ebay-details`
```
LuLaValva, line 58: This list is going to get _really_ long once everything's in here... maybe we should have a separate markdown file for each component?

docs/adr/0005-evo-react-child-component-composition.md
```diff
@@ -0,0 +1,48 @@
...
+
+```tsx
+<EvoDetails size="small" onToggle={handler}>
+  <EvoDetailsSummary>
+    <EvoDetailsLeading>
+      <EvoIconLightbulb16 />
+    </EvoDetailsLeading>
+    <EvoDetailsLabel>How do I get started?</EvoDetailsLabel>
+  </EvoDetailsSummary>
+  <EvoDetailsContent>Content</EvoDetailsContent>
```
LuLaValva, line 30: Are we planning on always having `*Content` tags? I'd hope that in many cases we can just keep the content inline
```tsx
<EvoDetails size="small" onToggle={handler}>
  <EvoDetailsSummary>
    <EvoDetailsLeading>
      <EvoIconLightbulb16 />
    </EvoDetailsLeading>
    How do I get started?
  </EvoDetailsSummary>
  Content
</EvoDetails>

File: packages/evo-marko/src/tags/evo-details/examples/with-leading.marko [hunk 4]
```diff
@@ -0,0 +1,5 @@
+<evo-details ...input>
+    <@summary>Details</@summary>
+    <@leading><evo-icon-lightbulb-16/></@leading>
```
LuLaValva, line 3: Can we nest `<@leading>` inside `<@summary>` instead? It'll require a `<const/{ leading, ...rest }=summary/>` on the consuming side but I think that's worth it

```suggestion
    <@summary>
      <@leading><evo-icon-lightbulb-16/></@leading>
      Details
    </@summary>
```

docs/adr/0005-evo-react-child-component-composition.md
```diff
@@ -0,0 +1,48 @@
...
+</EvoDetails>
```
LuLaValva, line 31: Thoughts on using an object instead of individual components, like Radix does? Might make more sense for grouping, but I'm not sure
```tsx
<EvoDetails.Root size="small" onToggle={handler}>
  <EvoDetails.Summary>
    <EvoDetails.Leading>
      <EvoIconLightbulb16 />
    </EvoDetails.Leading>
    <EvoDetails.Label>How do I get started?</EvoDetails.Label>
  </EvoDetails.Summary>
  <EvoDetails.Content>Content</EvoDetails.Content>
</EvoDetails.Root>
```

packages/evo-react/src/evo-details/context.ts
```diff
@@ -0,0 +1,13 @@
+import { createContext, useContext } from "react";
```
LuLaValva, line 1: If it was up to me I'd probably put all of the component parts in one file, you know better than I do though

LuLaValva (commented):
Looks great! Just one change on the Marko side and some high-level thoughts on design decisions for React

.claude/skills/evo-app-migrate-react/SKILL.md
HenriqueLimas: Yeah probably we should start moving around.

docs/adr/0005-evo-react-child-component-composition.md
HenriqueLimas: We can have both at some point like base-ui has.

packages/evo-marko/src/tags/evo-details/examples/with-leading.marko
HenriqueLimas: This will be wrong no? It will render the icon inside the summary text? But it is a sibling

packages/evo-react/src/evo-details/context.ts
HenriqueLimas: I think having in a separate file prevents the risk of circular dependencies since we have multiple components.

docs/adr/0005-evo-react-child-component-composition.md
HenriqueLimas: Not always, it will depend on the component and how skin organizes it. Having the component allow the developer to add other properties like class names to the container, but it will depend for each component might be different.

packages/evo-marko/src/tags/evo-details/examples/with-leading.marko
LuLaValva: In the markup it is a child of `<summary>`, not a sibling...

HenriqueLimas: yes but `@summary` is actually the `label` not the `details__summary` https://github.com/eBay/evo-web/pull/640/changes#diff-341fd61a09aec307e916b291902025382ead9dd66813cb13883cc8004c6119deR32

LuLaValva: Hmm, I guess you're right but it does mess with my mental model a little bit. I've also never been a fan of `@leading` as a name, maybe that's part of it. How about `leadingIcon`?

LuLaValva (approved):
Approving because `evo-react` is still in alpha, but some of the conversations here are maybe worth continuing as more components are migrated

Issue #616, opened 2026-04-09
ebay-section-notice: action link text is bolded


hongjilee-1:
### I verified there's no existing issue for this bug.

- [x] There are no existing issues

### Current behavior

Links and fake-link buttons in @cta slot end up having bolded text as shown in the storybook of ebayui-core@16.7.0 & skin@19.24.0

### Expected behavior

DS specs shows that text on action CTA links are unbolded.

https://playbook.ebay.com/design-system/components/section-notice

### Package and version

ebayui-core@16.7.0 & skin@19.24.0

### Steps to reproduce

```bash
1.
2.
3.
...
```

ArtBlue:
This was already fixed and released. Closing...

PR #648, opened 2026-04-30
fix(skin): unbold action CTA links in page-notice and section-notice

SinhSinhAn:
## Summary

Closes #615 and #616 in one shot ,  both report the same bug on adjacent skin components.

In `page-notice` and `section-notice`, links and `button.fake-link` elements inside the `@cta` slot render bold, but the [Playbook design spec](https://playbook.ebay.com/design-system/components/section-notice) calls for unbolded CTA action text.

## Why this happens

Both components have blanket rules that bold every link in the component, regardless of where it sits:

```scss
.page-notice a               { font-weight: bold; }
.page-notice button.fake-link { font-weight: bold; }
.section-notice a               { font-weight: bold; }
.section-notice button.fake-link { font-weight: bold; }
```

Those rules cover legitimate cases (a link in the body or title that should be bold) but they also catch the `__cta` slot links, which the spec wants unbolded.

## Fix

Add a scoped override after the blanket rules:

```scss
.page-notice__cta a,
.page-notice__cta button.fake-link {
    font-weight: normal;
}
```

(and the parallel `.section-notice__cta` rules.) The override and the blanket rule share the same specificity (`0,1,1`), so source order matters — I placed the override below the blanket rules in both files and documented why with an inline comment, since this trips up future maintainers.

Non-CTA links keep the existing bold weight, so the behaviour change is scoped strictly to the `__cta` slot.

## Diff

`page-notice.scss` and `section-notice.scss` each get one new rule block. Compiled output (`dist/`) regenerated via `npm run build` in `packages/skin/`:

```diff
 .page-notice button.fake-link:hover {
     color: var(--page-notice-color, var(--color-foreground-on-inverse));
 }
+.page-notice__cta a,
+.page-notice__cta button.fake-link {
+    font-weight: 400;
+}
```

(`font-weight: 400` is the numeric form `cssnano` emits for `normal`, matching the rest of the file's convention of `400` / `700`.)

## Test plan

- [x] `npm run build:css` clean (stylelint passes)
- [x] Source SCSS / emitted CSS rule order verified — override sits after the blanket bold rules
- [x] Manual scan of existing storybook stories (`packages/skin/src/sass/section-notice/stories/*.stories.js` and `page-notice/stories/*`): no story exercises a non-CTA link, so no permutation should regress
- [ ] Percy visual regression on the storybook stories — needs maintainer trigger
- [ ] RTL story — `font-weight` is direction-independent, no change expected

## Files

- `packages/skin/src/sass/page-notice/page-notice.scss` (+12)
- `packages/skin/src/sass/section-notice/section-notice.scss` (+12)
- `packages/skin/dist/page-notice/page-notice.css` (+4)
- `packages/skin/dist/section-notice/section-notice.css` (+4)
- `.changeset/notice-cta-link-weight.md` (+5, `@ebay/skin: patch`)

5 files, +48 lines, 0 deletions.

### DIFF
--- a/packages/skin/src/sass/page-notice/page-notice.scss
+++ b/packages/skin/src/sass/page-notice/page-notice.scss
@@ -87,6 +87,20 @@ span[role="region"].page-notice {
     );
 }
 
+/*
+ * Action CTA links / fake-link buttons render with normal weight per the
+ * design spec at https://playbook.ebay.com/design-system/components/page-notice.
+ * The blanket `.page-notice a` and `.page-notice button.fake-link` rules
+ * above bold every link in the component; this restores the unbolded
+ * weight inside the __cta slot only. Must come after the blanket rules
+ * because they share the same specificity (0,1,1) and source order
+ * decides. See #615.
+ */
+.page-notice__cta a,
+.page-notice__cta button.fake-link {
+    font-weight: normal;
+}
+
 /* Override default outline color to avoid contrast issues */
 /* with the darker background colors in page notices */
 /* For explicit focus only: the `focus-visible` instead of `focus` */
--- a/packages/skin/src/sass/page-notice/page-notice.scss
+++ b/packages/skin/src/sass/page-notice/page-notice.scss
@@ -51,7 +51,6 @@ span[role="region"].page-notice {
     );
 
     font-size: var(--font-size-body);
-    font-weight: bold;
 }
 
 /* force links with text having more than one word to remain on same line */
@@ -77,7 +76,6 @@ span[role="region"].page-notice {
     );
 
     font-size: var(--font-size-body);
-    font-weight: bold;
 }
 
 .page-notice button.fake-link:hover {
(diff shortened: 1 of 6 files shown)


packages/skin/src/sass/page-notice/page-notice.scss
```diff
@@ -87,6 +87,20 @@ span[role="region"].page-notice {
...
+ * Action CTA links / fake-link buttons render with normal weight per the
+ * design spec at https://playbook.ebay.com/design-system/components/page-notice.
+ * The blanket `.page-notice a` and `.page-notice button.fake-link` rules
+ * above bold every link in the component; this restores the unbolded
+ * weight inside the __cta slot only. Must come after the blanket rules
+ * because they share the same specificity (0,1,1) and source order
+ * decides. See #615.
+ */
+.page-notice__cta a,
+.page-notice__cta button.fake-link {
```
ArtBlue: This looks to be a fragile specificity contract.

The fix works solely because of CSS source order is both the blanket rule and the override share identical specificity. This is documented in comments, but it's still a code smell.

Any future rule added after the override that re-boldens links in this component would silently win.
Build tool reordering or SCSS partials being split could break this without warning.
A more robust fix would be to scope the original blanket rules away from __cta using :not(), or restructure to only bold links in specific slots (__title, __body) rather than globally:

// Instead of overriding, scope the original rule more precisely so an override is not even necessary:
.page-notice__title a,
.page-notice__body a,
.page-notice__title button.fake-link,
.page-notice__body button.fake-link {
  font-weight: bold;
}

This would eliminate the need for the override entirely.

packages/skin/src/sass/page-notice/page-notice.scss
```diff
@@ -87,6 +87,20 @@ span[role="region"].page-notice {
     );
 }
 
+/*
+ * Action CTA links / fake-link buttons render with normal weight per the
```
ArtBlue: It's not necessary and is probably preferable to avoid adding links in comments. Actually, this entire comment block is not necessary, especially after changing how we're specifying the bolded elements instead of the opposite - creating exclusions. 

I realize you probably used an AI to do this work and the comments were the decision of the LLM, but in this instance, it's more noise than is helpful.

ArtBlue (changes requested):

SinhSinhAn:
I'll restructure to scope the bold rule from the start instead of overriding it. I'll also hit the other CTA stories while I'm at it. Pushing the rewrite shortly, Thank you for helping me become better at this

ArtBlue (approved):
Looks better! Thank you!

PR #652, opened 2026-05-04
fix(ebay-menu-button): remove broken expand restoration logic

LuLaValva:
Recent updates to `makeup-expander` changed its behavior and exposed faulty logic that attempts to store/restore expanded state. In certain cases where a re-render was happening during a state change, the component was locked into an "expanded" state.

Removing this logic fixes the issue.

https://github.com/user-attachments/assets/3148364b-d3d4-48a3-a94d-79a630d82214


saiponnada (approved):
Cool, LGTM!

PR #653, opened 2026-05-05
feat(evo-react): add accordion component

HenriqueLimas:
- Fixes #

## Description

Adds the new `EvoAccordion` component to `@evo-web/react` with id-based compound component composition:

- Adds `EvoAccordion`, `EvoAccordionItem`, `EvoAccordionSummary`, `EvoAccordionLeading`, `EvoAccordionLabel`, and `EvoAccordionContent` exports.
- Supports controlled and uncontrolled single-open and multi-open modes using `string` / `string[]` ids.
- Adds Storybook docs/stories, browser interaction tests, SSR snapshot tests, README, and a changeset.
- Updates app migration skill docs to move per-component migration details into linked files.
- Renamed `@details` to `@item` so it better explain that it can support multiple items to avoid using multiple accordion components.

## Notes

No GitHub issue number was provided when creating this PR.

Verification:

- `npm run build -w packages/evo-react` passed.
- Full pre-push hook failed on an unrelated `packages/ebayui-core` carousel browser test (`then it emitted the scroll event` expected one event but received zero), so the branch was pushed with `--no-verify`.

## Screenshots

N/A

## Checklist


packages/evo-react/src/accordion/accordion.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+          {...args}
+          open={open}
+          onOpenChange={handleOpenChange}
+          a11yText="Accordion Menu"
+        >
+          <EvoAccordionItem id="0">
+            <EvoAccordionSummary>
+              <EvoAccordionLabel>Shipping and returns</EvoAccordionLabel>
+            </EvoAccordionSummary>
+            <p>{LOREM}</p>
```
LuLaValva: Looks like we use `EvoAccordionContent` in some places, and raw children of `EvoAccordionItem` in others. Is this intended? Do we need `EvoAccordionContent`?

packages/evo-react/src/accordion/index.ts
```diff
@@ -0,0 +1,14 @@
+export { EvoAccordion } from "./accordion";
+export { EvoAccordionItem } from "./accordion-item";
+export { EvoDetailsSummary as EvoAccordionSummary } from "../details/details-summary";
+export { EvoDetailsLeading as EvoAccordionLeading } from "../details/details-leading";
+export { EvoDetailsLabel as EvoAccordionLabel } from "../details/details-label";
+export { EvoDetailsContent as EvoAccordionContent } from "../details/details-content";
```
LuLaValva, line 6: Are these re-exports necessary? If we want the API to more similar to marko, it could make sense to use
```marko
EvoAccordion
  EvoAccordionDetails
    EvoDetailsSummary
    EvoDetailsContent

File: packages/evo-react/src/accordion/accordion.tsx [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+  if (isOpen) {
+    return id;
+  }
+
+  return currentOpen === id ? undefined : currentOpen;
+}
+
+export function EvoAccordion(props: MultipleAccordionProps): JSX.Element;
+export function EvoAccordion(props: SingleAccordionProps): JSX.Element;
+export function EvoAccordion(props: EvoAccordionProps) {
```
LuLaValva, line 38: I like this overloading capability, more precise than the generic we are using in Marko

LuLaValva (commented):
Overall looks pretty good! One major suggestion for API change, but otherwise I like how this feels to consumers like a thin wrapper around `<ul>` and `<details>`.

Alternatively, maybe we could explore using some kind of [details `name=`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details#name) pattern to get even closer to the platform...

packages/evo-react/src/accordion/accordion.tsx
HenriqueLimas: Yes! We started doing it with button since it can be a anchor or button depending on the `href`. It helps also on the `ref` typing.

packages/evo-react/src/accordion/index.ts
HenriqueLimas: My though here was that `EvoDetails` usage is an implementation detail, and it feel strange to use a different name or import from two different places.

```tsx
import { EvoAccordion } from "@evo-web/react/accordion"
import { EvoDetails } from "@evo-web/react/details"

<EvoAccordion>
  <EvoDetails />
</EvoAccordion>
```

In Marko `@details` is an attribute tag of `<evo-accordion>` but the consumer does't specifically uses `evo-details`.

packages/evo-react/src/accordion/accordion.stories.tsx
HenriqueLimas: That was intentional as I saw that skin in details also don't need to have the details__content but I will put it back.

HenriqueLimas:
@LuLaValva the `name` proposal is interesting. Although I think they have different purpose. In `details` the name it is used to group details together and have the open one at time behavior, I think that is different then the single open/close behavior of accordion. If we use `name` might be confusing with what the native `name` means. Happy to discuss though.

LuLaValva (approved):

PR #660, opened 2026-05-06
feat(evo-react): migrate ebay-avatar to evo-avatar

HenriqueLimas:
- Fixes #

## Description

Migrates `ebay-avatar` from `@ebay/ebayui-core-react` to `@evo-web/react` as `evo-avatar`.

- Adds `EvoAvatar` and `EvoAvatarImage` components under `packages/evo-react/src/avatar/`
- Uses `a11yText` prop (mapped to `aria-label` internally) aligned with the evo-marko component
- Fixes `isFit` logic: uses `aspectRatio > 4/3` (portrait OR wide landscape) matching evo-marko — the original used `< 4/3` which was a bug
- Imports `@ebay/skin/avatar.mjs` directly in the component (no global CSS required)
- Uses `EvoIconAvatarSignedOut` instead of the `<EbayIcon name="..." />` pattern
- React 19 native `ref`, named function declarations, no `forwardRef`, no default exports
- Adds `subcomponents: { EvoAvatarImage }` in the Storybook meta so autodocs renders a props table tab per sub-component
- Updates the `evo-migrate-react` skill with guidance on the `subcomponents` convention for future migrations
- Adds browser tests (`vitest-browser-react`) and SSR snapshot tests

## Notes

The `isFit` bug fix is a deliberate behaviour change from `ebayui-core-react`: wide landscape images (aspect ratio > 4/3) now correctly get the `avatar--fit` class, consistent with the Marko implementation.

## Screenshots

N/A — visual output is identical to the existing `ebay-avatar` component; Percy will confirm.

## Checklist

### DIFF
--- a/.changeset/silver-wolves-dance.md
+++ b/.changeset/silver-wolves-dance.md
@@ -1,0 +1,5 @@
+---
+"@evo-web/react": minor
+---
+
+feat(avatar): add EvoAvatar and EvoAvatarImage components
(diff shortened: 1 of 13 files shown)


LuLaValva (approved):
I don't love that we need to bring in context for an avatar component, but it makes sense. This looks good overall.

PR #661, opened 2026-05-07
feat(evo-marko): evo-combobox

LuLaValva:
## Description

- Add `<evo-combobox>` to `@evo-web/marko`

## Screenshots

image

### DIFF
--- a/packages/evo-marko/src/tags/evo-combobox/README.md
+++ b/packages/evo-marko/src/tags/evo-combobox/README.md
@@ -1,0 +1,56 @@
+# evo-combobox
+
+An accessible combobox input with an optional inline listbox of suggestions. Follows the [ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).
+
+## Usage
+
+```marko
+<evo-combobox name="search" placeholder="Search…">
+  <@option text="Option A"/>
+  <@option text="Option B"/>
+  <@option text="Option C"/>
+</evo-combobox>
+```
+
+## Props
+
+| Prop            | Type                      | Default       | Description                                                               |
+| --------------- | ------------------------- | ------------- | ------------------------------------------------------------------------- |
+| `value`         | `string`                  | —             | Current input value (two-way bindable via `:=`)                           |
+| `open`          | `boolean`                 | —             | Controls listbox visibility (two-way bindable via `:=`)                   |
+| `autocomplete`  | `"list" \| "none"`        | `"none"`      | `"list"` filters options to those matching the typed text                 |
+| `listSelection` | `"manual" \| "automatic"` | `"automatic"` | Whether arrow-key navigation writes the highlighted option into the input |
+| `floatingLabel` | `string`                  | —             | Floating label text                                                       |
+| `borderless`    | `boolean`                 | —             | Removes the control border                                                |
+| `fluid`         | `boolean`                 | —             | Stretches the combobox to fill its container                              |
+| `strategy`      | `"fixed" \| "absolute"`   | —             | Listbox positioning strategy                                              |
+| `disabled`      | `boolean`                 | —             | Disables the input                                                        |
+
+All other attributes are forwarded to the underlying `<input>` element.
+
+## Attribute tags
+
+### `@option`
+
+| Attr     | Type      | Description                                      |
+| -------- | --------- | ------------------------------------------------ |
+| `text`   | `string`  | Display text for the option (required)           |
+| `value`  | `string`  | Optional value; defaults to `text`               |
+| `sticky` | `boolean` | Always shown regardless of `autocomplete` filter |
+
+### `@button`
+
+Optional button rendered inside the control (e.g. clear or toggle button). Accepts all `<button>` attributes plus icon content.
+
+```marko
+<evo-combobox name="search">
+  <@button
+    aria-label="Clear"
+    onClick() {
+      /* … */
+    }>
+    <evo-icon-clear-16/>
+  </@button>
+  <@option text="Option A"/>
+</evo-combobox>
+```
--- a/packages/evo-marko/src/tags/evo-combobox/README.md
+++ b/packages/evo-marko/src/tags/evo-combobox/README.md
@@ -1,56 +1,16 @@
-# evo-combobox
+<h1 style='display: flex; justify-content: space-between; align-items: center;'>
+    <span>
+        evo-combobox
+    </span>
+    <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
+        DS v1.0.0
+    </span>
+</h1>
 
-An accessible combobox input with an optional inline listbox of suggestions. Follows the [ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).
+An accessible combobox input with an optional inline listbox of suggestions.
 
-## Usage
+## Examples and Documentation
 
-```marko
-<evo-combobox name="search" placeholder="Search…">
-  <@option text="Option A"/>
-  <@option text="Option B"/>
-  <@option text="Option C"/>
-</evo-combobox>
-```
-
-## Props
-
-| Prop            | Type                      | Default       | Description                                                               |
-| --------------- | ------------------------- | ------------- | ------------------------------------------------------------------------- |
-| `value`         | `string`                  | —             | Current input value (two-way bindable via `:=`)                           |
-| `open`          | `boolean`                 | —             | Controls listbox visibility (two-way bindable via `:=`)                   |
-| `autocomplete`  | `"list" \| "none"`        | `"none"`      | `"list"` filters options to those matching the typed text                 |
-| `listSelection` | `"manual" \| "automatic"` | `"automatic"` | Whether arrow-key navigation writes the highlighted option into the input |
-| `floatingLabel` | `string`                  | —             | Floating label text                                                       |
-| `borderless`    | `boolean`                 | —             | Removes the control border                                                |
-| `fluid`         | `boolean`                 | —             | Stretches the combobox to fill its container                              |
-| `strategy`      | `"fixed" \| "absolute"`   | —             | Listbox positioning strategy                                              |
-| `disabled`      | `boolean`                 | —             | Disables the input                                                        |
-
-All other attributes are forwarded to the underlying `<input>` element.
-
-## Attribute tags
-
-### `@option`
-
-| Attr     | Type      | Description                                      |
-| -------- | --------- | ------------------------------------------------ |
-| `text`   | `string`  | Display text for the option (required)           |
-| `value`  | `string`  | Optional value; defaults to `text`               |
-| `sticky` | `boolean` | Always shown regardless of `autocomplete` filter |
-
-### `@button`
-
-Optional button rendered inside the control (e.g. clear or toggle button). Accepts all `<button>` attributes plus icon content.
-
-```marko
-<evo-combobox name="search">
-  <@button
-    aria-label="Clear"
-    onClick() {
-      /* … */
-    }>
-    <evo-icon-clear-16/>
-  </@button>
-  <@option text="Option A"/>
-</evo-combobox>
-```
+- [Storybook](https://ebay.github.io/evo-web/ebayui-core/?path=/story/form-input-evo-combobox)
+- [Storybook Docs](https://ebay.github.io/evo-web/ebayui-core/?path=/docs/form-input-evo-combobox)
+- [Code Examples](https://github.com/eBay/evo-web/tree/main/packages/evo-marko/src/tags/evo-combobox/examples)
--- a/packages/evo-marko/src/tags/evo-combobox/combobox.stories.ts
+++ b/packages/evo-marko/src/tags/evo-combobox/combobox.stories.ts
@@ -1,0 +1,130 @@
+import { buildExtensionTemplate } from "../../common/storybook/utils";
+import { type Meta } from "@storybook/marko";
+import Readme from "./README.md";
+import Component, { type Input } from "./index.marko";
+import DefaultTemplate from "./examples/default.marko";
+import DefaultCode from "./examples/default.marko?raw";
+import ControllableTemplate from "./examples/controllable.marko";
+import ControllableCode from "./examples/controllable.marko?raw";
+import AsyncFilteringTemplate from "./examples/async-filtering.marko";
+import AsyncFilteringCode from "./examples/async-filtering.marko?raw";
+
+export default {
+    title: "form input/evo-combobox",
+    component: Component,
+    parameters: {
+        docs: {
+            description: { component: Readme },
+        },
+    },
+    argTypes: {
+        autocomplete: {
+            type: "string",
+            options: ["none", "list"],
+            control: "inline-radio",
+            description:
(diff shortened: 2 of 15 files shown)


packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,216 @@
...
+    listSelection = "automatic",
+    floatingLabel,
+    strategy,
+    option: options = [],
+    button,
+    fluid,
+    onBlur,
+    onFocus,
+    onKeyDown,
+    placeholder,
```
HenriqueLimas, line 43: Do we need to remove `open` and `openChange` as we did in the other component

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,216 @@
...
+    ...htmlInput
+}=input>
+
+<id/inputId=id>
+<let/value:=input.value>
+<let/open:=input.open>
+<let/activeIndex=(-1)>
+<let/displayValue=(null as string | null)>
```
HenriqueLimas: Is this also a valid Marko syntax? I think using `as` it is usually not typesafe


```suggestion
<let/displayValue: string | null =null>
```

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,216 @@
...
+                        open = false;
+                    } else if (e.key === "Escape") {
+                        open = false;
+                        activeIndex = -1;
+                    }
+                    onKeyDown && onKeyDown(e, target);
+                }>
+            <if=button>
+                <button
+                    ...button
```
HenriqueLimas, line 155: Can't we use the `evo-icon-button` component here instead or it is not necessary?

packages/evo-marko/src/tags/evo-combobox/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```marko
+<evo-combobox name="search" placeholder="Search…">
+  <@option text="Option A"/>
+  <@option text="Option B"/>
+  <@option text="Option C"/>
+</evo-combobox>
+```
+
+## Props
```
HenriqueLimas: In `evo-react` we are moving away from documentation in the README to storybook only and link to storybook on the readme. Reason is that it is hard to maintain this list updated most of the time. In evo-marko we are using the README as the documentation, but we also have the `args` from storybook so it will duplicate information I think.

packages/evo-marko/src/tags/evo-combobox/style.ts
```diff
@@ -0,0 +1,3 @@
+import "@ebay/skin/combobox";
+import "@ebay/skin/icon-button";
+import "@ebay/skin/floating-label";
```
HenriqueLimas: This one shouldn't be necessary now that is imported by evo-floating-label component


```suggestion
```

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: _Great_ catch, yes we do need this.

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: Unfortunately our language server doesn't do well with `: foo` on tag variables 🙈. I agree it looks nicer to do what you say but for now we need to keep `as`

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: I think icon button adds extra functionality that we don't need, we didn't use it for `<ebay-combobox>`, but I don't know why not.

It's probably a good idea to unify here, especially because it will help to enforce `a11yText` (which I forgot in my controllable example, so it already fixed one bug 😳)

packages/evo-marko/src/tags/evo-combobox/README.md
LuLaValva: Oh, yep. This is what I get for not looking at AI's README

packages/evo-marko/src/tags/evo-combobox/style.ts
LuLaValva: Right! Thanks.

packages/evo-marko/src/tags/evo-combobox/test/test.browser.ts
```diff
@@ -0,0 +1,165 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { render, fireEvent, cleanup, waitFor } from "@marko/testing-library";
+import { composeStories } from "@storybook/marko";
+import * as stories from "../combobox.stories";
+
```
ArtBlue, line 5: **Blocking:** `Filtering` is destructured here but is not exported from `combobox.stories.ts`. The stories file exports `Default`, `Controllable`, and `AsyncFiltering`. Either rename this to `AsyncFiltering`, or add a dedicated `Filtering` story (e.g. `Default` with `autocomplete="list"` and inline options) so the test suite can actually run.

packages/evo-marko/src/tags/evo-combobox/test/test.server.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, it } from "vitest";
+import { composeStories } from "@storybook/marko";
+import { snapshotHTML } from "../../../common/test-utils/snapshots";
+import * as stories from "../combobox.stories";
+
```
ArtBlue, line 5: **Blocking:** Both `Filtering` and `FloatingLabel` are destructured here but neither is exported from `combobox.stories.ts`. These will resolve to `undefined` and cause `composeStories` to throw at runtime, failing the entire SSR test suite. Add the missing story exports.

packages/evo-marko/src/tags/evo-combobox/combobox.stories.ts
```diff
@@ -0,0 +1,130 @@
...
+    ControllableTemplate,
+    ControllableCode,
+    { "aria-label": "Campaign", placeholder: "Choose a campaign" },
+);
+
+export const AsyncFiltering = buildExtensionTemplate(
+    AsyncFilteringTemplate,
+    AsyncFilteringCode,
+    { placeholder: "Type a country name..." },
+);
```
ArtBlue, line 130: **Blocking:** Both test files (`test.browser.ts` and `test.server.ts`) expect `Filtering` and `FloatingLabel` story exports that are not defined here. Missing at minimum:
- `Filtering` — a story with `autocomplete="list"` and a set of inline `@option` items (so the filtering test can assert on option count after typing)
- `FloatingLabel` — a story with a `floatingLabel` prop set

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,219 @@
...
+                        ...htmlOption
+                        id=`${inputId}-option-${i}`
+                        role="option"
+                        class=[
+                            "combobox__option",
+                            i === activeIndex && "combobox__option--active",
+                            optionClass,
+                        ]
+                        tabindex="-1"
+                        aria-selected=((optionValue || text) === value) && "true"
```
ArtBlue: **Blocking (a11y):** `aria-selected` is only set to `"true"` for the selected option and omitted for all others. The WAI-ARIA spec marks `aria-selected` as a required attribute on `role="option"` inside a `role="listbox"`, and screen readers rely on the explicit `"false"` to distinguish "not selected" from "state unknown". Change to:
```
aria-selected=((optionValue || text) === value ? "true" : "false")
```

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,219 @@
...
+                    }
+                    onBlur && onBlur(e, target);
+                }
+                onInput(e, target) {
+                    value = target.value;
+                    displayValue = null;
+                    activeIndex = -1;
+                    if (input.open !== false) open = true;
+                }
+                onKeyDown(e: KeyboardEvent, target: HTMLInputElement) {
```
ArtBlue: **Non-blocking:** Project convention (CLAUDE.md `correctness_guards`) says not to annotate `e`/`target` on native HTML tag handlers — types are inferred. The other handlers (`onFocus`, `onBlur`, `onInput`) correctly omit annotations. Remove the explicit types:
```
onKeyDown(e, target) {
```
Note: the `onFocusOut` annotations on the dynamic `<${fluid ? "div" : "span"}>` tag at line 72 are correct and required since dynamic tags have no inferred type info.

ArtBlue (commented):
Claude PR Review. Full list here: https://github.com/eBay/evo-web/tasks/fe6604f5-9c3b-4888-9e89-e3482af049e7

Three blocking issues that will cause test suite failures and a WCAG violation: both test files import stories that don't exist in the stories file (`Filtering`, `FloatingLabel`), and non-selected listbox options are missing the required `aria-selected` attribute. One non-blocking type annotation nit.

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,216 @@
...
+        "type" | "role" | "aria-autocomplete" | "aria-haspopup" | "aria-expanded"
+    > {
+    autocomplete?: "list" | "none";
+    listSelection?: "manual" | "automatic";
+    floatingLabel?: string;
+    borderless?: boolean;
+    fluid?: boolean;
+    strategy?: "fixed" | "absolute";
+    open?: boolean;
+    openChange?: (open: boolean) => void;
```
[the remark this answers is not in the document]
LuLaValva: It is used by `:=`, read the Marko docs next time

ArtBlue (commented):
Previous blocking issues are all resolved — great cleanup. One new blocking bug introduced by the `tempValue` refactor: Escape key doesn't clear `tempValue`, so keyboard-navigated values commit on blur. Three non-blocking observations: redundant `aria-owns`, always-in-DOM listbox relying on skin CSS, and lost SSR coverage.

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,217 @@
...
+                            document.getElementById(`${inputId}-option-${activeIndex}`)?.scrollIntoView({ block: "nearest" });
+                        }
+                    } else if (e.key === "Enter") {
+                        if (open && activeIndex >= 0) {
+                            e.preventDefault();
+                            value = visibleOptions[activeIndex].text;
+                        }
+                        open = false;
+                    } else if (e.key === "Escape") {
+                        open = false;
```
ArtBlue, line 147: **Blocking:** Escape closes the listbox but doesn't clear `tempValue`. Repro: arrow-navigate to an option (sets `tempValue`), press Escape (intending to cancel), then click away — `onBlur` sees `tempValue` is still set and commits it as the new value, overriding the user's intent to cancel.

Add `tempValue = null` here:
```
} else if (e.key === "Escape") {
    open = false;
    activeIndex = -1;
    tempValue = null;  // discard keyboard navigation on cancel
}
```

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,217 @@
...
+            <input
+                ...htmlInput
+                id=inputId
+                type="text"
+                role="combobox"
+                disabled=disabled
+                value:=value
+                placeholder=showPlaceholder && placeholder
+                autocomplete="off"
+                aria-owns=listboxId
```
ArtBlue, line 98: **Non-blocking:** `aria-owns` alongside `aria-controls` referencing the same element is redundant. The ARIA 1.2 combobox pattern specifies `aria-controls` to reference the popup listbox — `aria-owns` is only needed when the DOM parent-child relationship doesn't match the desired accessibility tree structure. Since the listbox is already a descendant of the same combobox widget, `aria-controls` alone is sufficient. Consider removing `aria-owns`.

packages/evo-marko/src/tags/evo-combobox/index.marko
```diff
@@ -0,0 +1,217 @@
...
+                <evo-icon-button
+                    ...button
+                    onMouseDown(e, el) {
+                        // prevent input blur when clicking the action button
+                        e.preventDefault();
+                        button.onMouseDown && button.onMouseDown(e, el);
+                    }/>
+            </if>
+        </span>
+        <div
```
ArtBlue, line 163: **Non-blocking:** The listbox is now always in the DOM (the previous `<if=open>` guard was removed). AT visibility when the combobox is closed now depends entirely on the skin CSS using `display: none` for `.combobox__listbox` when `.combobox--expanded` is absent — `display: none` correctly removes the element from the accessibility tree; `visibility: hidden` or `opacity: 0` would not. Worth confirming the skin CSS uses `display` for this toggle.

packages/evo-marko/src/tags/evo-combobox/test/test.server.ts
```diff
@@ -0,0 +1,32 @@
...
+    it("renders fluid", async () => {
+        await snapshotHTML(Default, { fluid: true });
+    });
+
+    it("renders with pre-selected value", async () => {
+        await snapshotHTML(Default, { value: "Basic Offer" });
+    });
+
+    it("renders with fixed strategy", async () => {
+        await snapshotHTML(Default, { strategy: "fixed" });
```
ArtBlue, line 30: **Non-blocking:** The previous revision had SSR snapshots for `floatingLabel` and `Filtering` (autocomplete='list') modes — they were removed here rather than replaced. The snapshot count is now 6 cases vs. the prior 8. Consider adding:
```ts
it("renders floating label", async () => {
    await snapshotHTML(Default, { floatingLabel: "Campaign" });
});
it("renders filtering variant", async () => {
    await snapshotHTML(Default, { autocomplete: "list" });
});
```

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: Nice catch

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: Skin uses `display: none`, this is the preferred behavior

packages/evo-marko/src/tags/evo-combobox/index.marko
LuLaValva: This is consistent with MIND patterns— if we update we should do so there first

packages/evo-marko/src/tags/evo-combobox/test/test.browser.ts
```diff
@@ -0,0 +1,165 @@
...
+                        "aria-expanded",
+                        "false",
+                    );
+                });
+            });
+
+            describe("when an option is clicked", () => {
+                beforeEach(async () => {
+                    const options = component.getAllByRole("option");
+                    await fireEvent.mouseDown(options[2]);
```
HenriqueLimas, line 113: In react we prefer to use userEvent rather than fireEvent, in that way you don't need to trigger mouseDown and the click. That will be already fired by user event. It is more real user, wondering though if marko has some limitations on using (I don't think it should)

HenriqueLimas (approved):

PR #663, opened 2026-05-07
feat(evo-react): add `as` prop to EvoButton for custom link component support

HenriqueLimas:
- Fixes #

## Description

Adds an `as` prop to `EvoButton`'s anchor variant (i.e. when `href` is provided), following the same pattern established in `evo-breadcrumbs`. This allows consumers to substitute the native `<a>` element with any component that accepts the same props — such as React Router's `<Link>`, Next.js's `<Link>`, or any other framework router.

**Usage example:**
```tsx
import { Link } from 'react-router';

function LinkHref({ href, ...rest }) {
  return <Link to={href} {...rest} />
}

<EvoButton href="/home" as={LinkHref} priority="primary">
  Go Home
</EvoButton>
```

**Changes:**
- `types.ts` — Added `as?: ComponentType<ComponentProps<"a">>` to `AnchorButtonProps` and `as?: never` to `NativeButtonProps` (so the prop is safely destructurable across the union without TypeScript errors)
- `button.tsx` — Destructures `as` from props (keeping it out of `...rest` so it's never leaked to the DOM) and uses the resolved `Component` in the anchor branch
- `button.stories.tsx` — Widened `Meta` type to `AnchorButtonProps` and documented `as` in `argTypes`
- `test/test.server.tsx` — Added SSR snapshot test for the `as` prop
- `test/test.browser.tsx` — Added two browser tests: verifies custom component renders with its own attributes, and that all `fake-btn` classes are correctly applied

## Notes

- `as` is **only** valid when `href` is also provided. Passing `as` without `href` is a TypeScript error (`as?: never` on `NativeButtonProps`)
- The prop is extracted before `...rest` spread so it is never forwarded to the DOM
- Pattern is identical to `evo-breadcrumbs` (`BreadcrumbLink` component on the `evo-react-breadcrumb` branch)

## Screenshots

N/A — no visual change; the rendered HTML is identical to the default `<a>` unless a custom component changes its own output.

## Checklist

### DIFF
--- a/.changeset/evo-button-as-prop.md
+++ b/.changeset/evo-button-as-prop.md
@@ -1,0 +1,5 @@
+---
+"@evo-web/react": patch
+---
+
+Add `as` prop to `EvoButton` anchor variant to support custom link components (e.g. React Router `Link`, Next.js `Link`). Only applies when `href` is provided.
--- a/packages/evo-react/src/button/button.stories.tsx
+++ b/packages/evo-react/src/button/button.stories.tsx
@@ -2,6 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react-vite";
 import { EvoButton } from "./button";
 import { EvoButtonCell } from "./button-cell";
 import type { AnchorButtonProps } from "./types";
+import { ComponentProps } from "react";
 
 // Use AnchorButtonProps so `as` (anchor-only prop) is a valid argType key.
 const meta: Meta<AnchorButtonProps> = {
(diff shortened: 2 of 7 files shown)


LuLaValva (dismissed):
Nice!

HenriqueLimas:
@LuLaValva maybe I should call it `is` instead, to be similar to HTML spec https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/is, what do you think?

LuLaValva (approved):
I like `is` more, it fits HTML spec better and it's also not a reserved word! Let's move to that universally, in `evo-react` and `evo-marko`. Maybe we should merge this first and then do a follow-up PR that moves them all to `is`.

HenriqueLimas:
@LuLaValva maybe the only issue is if the developer wants to actually use the native `is` property.

PR #672, opened 2026-05-12
feat: improve context-auditor skill score from 58% to 85%

yogesh-tessl:
Hey @ArtBlue 👋

I ran your skills through `tessl skill review` at work and found some targeted improvements for your `context-auditor` skill. Here's the before/after:

| Skill | Before | After | Change |
|-------|--------|-------|--------|
| context-auditor | 58% | 85% | **+27%** |

picked `context-auditor` because it had the most improvement headroom at 58%, the description was missing explicit trigger guidance and the content had significant redundancy that hurt conciseness and progressive disclosure scores.

<details>
<summary>Changes made</summary>

- **Improved frontmatter description** - added explicit "Use when..." clause and expanded trigger terms (`system prompt`, `AI instructions`, `prompt engineering files`) so Claude selects this skill more reliably
- **Switched to third-person voice** - replaced "You are a Principal Context Architect" with a neutral description-style opening
- **Removed redundant sections** - "Prioritize Fidelity Over DRY" and "Discovery-First Architecture" were restating principles already covered in the Audit Criteria; anti-pattern/good-pattern lists also duplicated criteria content
- **Extracted to REFACTORING_GUIDE.md** - moved the removed content into a companion reference file for progressive disclosure, keeping the main skill lean while preserving all domain knowledge
- **Consolidated refactoring mandates** - replaced verbose code examples with concise inline summaries; code examples live in the reference file
- **Added validation step** - workflow now includes an explicit "re-score to verify improvement" step before presenting results
- **Consolidated checklists** - merged single-file and repo-wide checklists into one, removing duplicate items

</details>
---

Honest disclosure. I work at https://github.com/tesslio where we build tooling around skills like these. Not a pitch, just saw room for improvement and wanted to contribute.

If you want to self-improve your skills, or define your own scenarios to pressure test, just ask your agent (Claude Code, Codex, etc.) to evaluate and optimize your skill with Tessl. Ping me [@yogesh-tessl](https://github.com/yogesh-tessl), if you hit any snags.

### DIFF
--- a/.claude/skills/context-auditor/REFACTORING_GUIDE.md
+++ b/.claude/skills/context-auditor/REFACTORING_GUIDE.md
@@ -1,0 +1,40 @@
+# Context Auditor — Refactoring Guide
+
+Extended patterns and checklists for the [context-auditor](SKILL.md) skill.
+
+## Fidelity Over DRY
+
+- Verbose, redundant configs are GOOD if fully cached
+- WET (Write Everything Twice) prevents attention decay
+- 5,000-token config at 90% cache rate > 1,100-token "clean" config
+
+## Discovery-First Architecture
+
+- ❌ **DON'T include:** File structures, directory layouts, naming patterns visible via file exploration
+- ✅ **DO include:** Non-obvious conventions, version-specific syntax, framework API differences
+- **Test:** "Could an AI discover this with Read/Glob/Grep in 1-2 files?" → If yes, replace with reference
+
+## Anti-Patterns to Flag
+
+- Frequently-used material extracted to skills (breaks caching)
+- Critical rules buried mid-file (lost in the middle)
+- No XML encapsulation for Claude (context bleeding)
+- Missing Correctness Guards (version hallucination risk)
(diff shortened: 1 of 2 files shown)


ArtBlue:
Hey @yogesh-tessl , thanks so much for the contribution! 

This skill is the result of a thorough, research-backed set of guidance. It was sourced by _how_ the system prompt works and less by conventions that have become commonplace, some of which are deeply flawed. 

This change actually breaks that and in some ways moves things backwards. It also moves the guidance backwards to a more generic approach rather than something that is tailored specifically for our system prompt. Our skill is not meant to be a generalized solution, but a way to maintain high fidelity of our own system prompt.

That system prompt was created by reverse engineering the guidance for the most optimal system prompt. The skill was reverse engineered from the optimal system prompt. Some smaller changes in this PR do seem to be good adjustments and we'll definitely apply them soon, but the changes to the skill overall will degrade out system prompt, so I'll have to close this for now.

P.S. Please don't hesitate to contribute things in the future, but we do prefer an issue to be created first before PRs, so we can first triage the problem statement.

PR #674, opened 2026-05-12
feat(evo-react): migrate ebay-icon-button to evo-icon-button

HenriqueLimas:
- Fixes #

## Description

Migrates `ebay-icon-button` from `@ebay/ebayui-core-react` to a new `EvoIconButton` component in `@evo-web/react`.

**New files in `packages/evo-react/src/icon-button/`:**
- `icon-button.tsx` — main component (named function, React 19 native ref, no `forwardRef`)
- `types.ts` — all exported types
- `index.ts` — named re-exports
- `icon-button.stories.tsx` — Storybook stories co-located with source
- `test/test.browser.tsx` — browser interaction tests (`vitest-browser-react`)
- `test/test.server.tsx` — SSR snapshot tests
- `README.md`

**Key API differences vs `ebay-icon-button`:**
- Icon passed as `children` instead of an `icon` prop — aligns with the evo-marko content slot pattern
- `aria-label` renamed to `a11yText` (`string | null`) — mapped internally to `aria-label`
- `badgeAriaLabel` renamed to `badgeA11yText`
- `forwardedRef` dropped — use the native `ref` prop (React 19)
- `as` prop added on anchor variant — supports custom link components (e.g. React Router `Link`), matching `EvoButton`
- `partiallyDisabled` prop added — renders `aria-disabled="true"` while keeping the element interactive
- Skin CSS bundled automatically via `.mjs` imports (`icon-button`, `button`, `link`)

Also adds a patch changeset and updates the `evo-app-migrate-react` skill with a linked `components/evo-icon-button.md` migration guide.

## Notes

N/A

## Screenshots

N/A — no visual changes to existing components; new component matches existing `ebay-icon-button` skin output.

## Checklist


packages/evo-react/src/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,117 @@
...
+    onKeyDown?.(
+      event as KeyboardEvent<HTMLButtonElement> &
+        KeyboardEvent<HTMLAnchorElement>,
+    );
+    if (event.key === "Escape" && !disabled) {
+      onEscape?.(
+        event as KeyboardEvent<HTMLButtonElement> &
+          KeyboardEvent<HTMLAnchorElement>,
+      );
+    }
```
LuLaValva, line 67: Do we still need `onEscape`? Part of the `evo-*` effort is to use our "breaking changes" opportunity to remove unnecessary bloat, we could expect that people migrate to do this check themselves in `onKeyDown`

packages/evo-react/src/icon-button/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  partiallyDisabled?: boolean;
+  badgeNumber?: number;
+  /**
+   * The accessible label of the badge.
+   *
+   * English default to be overridden is `""`
+   *
+   * Pass `null` explicitly _only_ if alternative accessibility information is present.
+   */
+  badgeA11yText?: string | null;
```
LuLaValva: Seems like this is documented as optional but also not optional? If the default is `""` and we have `?:` then it looks like other components just use
```suggestion
  badgeA11yText?: string;
```
and add minimal documentation in the storybook. I imagine in many cases the number itself suffices for a11y text

LuLaValva (dismissed):

packages/evo-react/src/icon-button/icon-button.tsx
HenriqueLimas: This is currently aligned with evo-marko, we can remove from both in a separate PR and maybe provide a helper function for this

Issue #675, opened 2026-05-13
Add new core and semantic tokens to the system


dlacefield:
### Description

**Tokens added by DT:**

- color.core.yellow.550
- color.background.warning
- color.foreground.warning

Helpful links:

**Playbook tokens pages**
[Filtered to Color > Core > Yellow](https://playbook.ebay.com/design-system/tokens/color-tokens?tab=core&q=yellow)
[Filtered to Color > Semantic > Warning](https://playbook.ebay.com/design-system/tokens/color-tokens/?tab=semantic&q=warning)

**Web tokens:**
Version 2.3.4 of the @ebay/design-tokens [NPM package](https://www.npmjs.com/package/@ebay/design-tokens) has all the new tokens

### Engineering pre-requisites

- [ ] Supports dark mode.
- [ ] Supports responsive design.
- [ ] Tokens have been provided (or already exist).

ArtBlue:
We already implemented this.

PR #678, opened 2026-05-14
feat(dependabot): track _all_ updates for design-tokens package

LuLaValva:
Configure dependabot so that it triggers a PR for all patch updates to `@ebay/design-tokens`, to ensure that we are always up to date.
- This should remove the need for issues like #675


ArtBlue (approved):
Looks good.

PR #679, opened 2026-05-14
feat(evo-marko): evo-calendar

LuLaValva:
- Add tags API `<evo-calendar>` component

### API Changes `ebay-calendar` → `evo-calendar`

- `navigable` + `get-a11y-show-month-text` → `a11yNavigateText`
  - callback's presence necessary for navigation
- `interactive` + `range` → `selectMode?: "day" | "range"`
  - Not 100% sure about this one day & range work nicely with controllable `selected` but we might need to add a third `"custom"` option in case devs just want days to be buttons
- `selected: DayISO | [DayISO, DayISO]` → `selected: DayISO | { from: DayISO, to: DayISO }` + `selectedChange`
  - Maybe unnecessary, but it simplified some logic and made it easier to read
- `disable-*` → `disable: { * }`
  - Also added `callback` for completely custom disable schemes
- `a11y-range-*-text` → `a11yRangeText: { * }`
  - Allows us to be a little more strict with TS
- `a11y-separator` → (removed)
  - In other places we unconditionally use `" - "` so this is for consistency

### New in `evo-calendar`

- `viewStart` + `viewStartChange`
  - Force the calendar to show a specific month or set of months, defaults to the month containing `today`

### Notes

- For this first pass, I didn't ever force `a11y` text. I think this can be improved in a variety of ways
  - `a11yTodayText` and `a11ySelectedText` are only used in non-interactive dialogs, and required in that case
  - `a11yRangeText` should be required for `selectMode: "range"`
- CLDR updated its structure, so we needed to change our process to convert it to locale info


packages/evo-marko/src/tags/evo-calendar/test/test.browser.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        it("does not render navigation buttons when getA11yNavigateText is absent", () => {
+            expect(component.queryAllByRole("button")).toHaveLength(0);
+        });
+    });
+
+    describe("given an interactive calendar", () => {
+        beforeEach(async () => {
+            component = await render(Default, {
+                interactive: true,
```
ArtBlue: **Blocking:** `interactive` is not a property in the new `Input` interface. Use `selectMode: "day"` (or `"range"`) instead. Every occurrence of `interactive: true` and `interactive: false` throughout this file needs to be updated.

packages/evo-marko/src/tags/evo-calendar/test/test.browser.ts
```diff
@@ -0,0 +1,125 @@
...
+                todayISO: "2025-01-15",
```
ArtBlue: **Blocking:** `todayISO` is not a valid prop. The new API uses `today`. Rename all occurrences in this file (lines 35, 62–63, 77, 94, 111).

packages/evo-marko/src/tags/evo-calendar/test/test.browser.ts
```diff
@@ -0,0 +1,125 @@
...
+            expect(selected).toHaveAttribute("aria-pressed", "true");
+        });
+    });
+
+    describe("given a navigable calendar", () => {
+        beforeEach(async () => {
+            component = await render(Default, {
+                interactive: true,
+                todayISO: "2025-01-15",
+                getA11yNavigateText: (month: string) => `Show ${month}`,
```
ArtBlue: **Blocking:** `getA11yNavigateText` is not a valid prop. The new API is `a11yNavigateText`, and it takes two parameters: `(monthName: string, dir: "next" | "prev") => string`. Update the callback signature accordingly, e.g. `` (month, dir) => `Show ${month}` ``.

packages/evo-marko/src/tags/evo-calendar/test/test.browser.ts
```diff
@@ -0,0 +1,125 @@
...
+            expect(buttons.length).toBeGreaterThan(2);
+        });
+    });
+
+    describe("given a calendar with disabled dates", () => {
+        beforeEach(async () => {
+            component = await render(Default, {
+                interactive: true,
+                todayISO: "2025-01-15",
+                disabled: { list: ["2025-01-10", "2025-01-20"] },
```
ArtBlue: **Blocking:** `disabled` is not a valid prop. The new API uses `disable` (e.g. `disable: { list: ["2025-01-10", "2025-01-20"] }`).

packages/evo-marko/src/tags/evo-calendar/test/test.browser.ts
```diff
@@ -0,0 +1,125 @@
...
+            expect(disabled.length).toBeGreaterThanOrEqual(2);
+        });
+    });
+
+    describe("given a non-interactive calendar with getHref", () => {
+        beforeEach(async () => {
+            component = await render(Default, {
+                interactive: false,
+                todayISO: "2025-01-15",
+                getHref: (iso: string) =>
```
ArtBlue: **Blocking:** `getHref` is not a valid prop. The new API uses `linkBuilder: (iso: DayISO) => string | false | null | undefined`.

packages/evo-marko/src/tags/evo-calendar/test/test.server.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const { Default, Controllable } = composeStories(stories);
+
+describe("evo-calendar SSR", () => {
+    it("renders default display calendar", async () => {
+        await snapshotHTML(Default);
+    });
+
+    it("renders with todayISO pinned", async () => {
+        await snapshotHTML(Default, { todayISO: "2025-01-15" });
```
ArtBlue: **Blocking:** `todayISO` is not a valid prop — use `today`. Also applies to lines 20, 27, 35, 43, 51, 58.

packages/evo-marko/src/tags/evo-calendar/test/test.server.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    it("renders interactive mode", async () => {
+        await snapshotHTML(Default, {
+            interactive: true,
```
ArtBlue: **Blocking:** `interactive` is not a valid prop — use `selectMode: "day"`. Applies to all occurrences throughout this file.

packages/evo-marko/src/tags/evo-calendar/test/test.server.ts
```diff
@@ -0,0 +1,62 @@
...
+            todayISO: "2025-01-15",
+            selected: { from: "2025-01-05", to: "2025-01-12" },
+        });
+    });
+
+    it("renders with navigable header", async () => {
+        await snapshotHTML(Default, {
+            interactive: true,
+            todayISO: "2025-01-15",
+            getA11yNavigateText: (month: string) => `Show ${month}`,
```
ArtBlue: **Blocking:** `getA11yNavigateText` → `a11yNavigateText` (same issue as browser test).

ArtBlue (commented):
Claude Code Review...

Good work adding `evo-calendar` — the API redesign is cleaner and the keyboard navigation is well thought out. There are **9 blocking issues** to fix before merge:

1. **All test files use old prop names** that don't match the new `Input` interface (`interactive`, `todayISO`, `getA11yNavigateText`, `disabled`, `getHref`). Every test case will either receive unknown props and test nothing, or fail outright.
2. **`<table>` is missing `role="grid"`** — the browser test explicitly queries `getByRole("grid")`, which will fail because a plain `<table>` has the implicit ARIA role `table`, not `grid`.
3. **`a11yTodayText` JSDoc has the wrong English default** (`"selected"` should be `"today"`).

Two non-blocking accessibility notes are also inline below.

packages/evo-marko/src/tags/evo-calendar/test/test.server.ts
```diff
@@ -0,0 +1,62 @@
...
+        });
+    });
+
+    it("renders with disabled dates", async () => {
+        await snapshotHTML(Default, {
+            interactive: true,
+            todayISO: "2025-01-15",
+            disabled: { before: "2025-01-05", after: "2025-01-25" },
```
ArtBlue: **Blocking:** `disabled` → `disable` (same issue as browser test).

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * Must be present to allow navigation between months
+   *
+   * Should return a string like `"Show ${month}"`
+   */
+  a11yNavigateText?: (monthName: string, dir: "next" | "prev") => string;
+  /**
+   * Clipped text to communicate the current day cells to screen readers, required for non-interactive calendars
+   *
+   * English default to be overridden is `"selected"`
```
ArtBlue: **Blocking:** The JSDoc for `a11yTodayText` says the English default is `"selected"`, but it should be `"today"`. `a11ySelectedText` is the one that defaults to `"selected"`. This looks like a copy-paste error.

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -0,0 +1,366 @@
...
+          e.preventDefault();
+          const { year: veY, month: veM } = getMonthInfo(viewStart, numMonths - 1);
+          focus = toISO(new Date(Date.UTC(veY, veM + 1, 0)));
+          break;
+        }
+      }
+    }>
+    <for|i| until=numMonths>
+      <div class="calendar__month">
+        <table>
```
ArtBlue, line 211: **Blocking:** `<table>` is missing `role="grid"`. A plain `<table>` has the implicit ARIA role `table`, not `grid`, so the browser test's `getByRole("grid")` call will fail. Add `role="grid"` here. Per the ARIA APG grid pattern, `<td>` cells inside a grid should also carry `role="gridcell"` explicitly.

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -0,0 +1,366 @@
...
+          <caption>${getMonthName(locale, viewStart, i)}</caption>
+          <thead>
+            <tr>
+              <for|dayName| of=weekdayLabels>
+                <th>${dayName}</th>
```
ArtBlue, line 216: **Non-blocking:** `<th>` column headers should carry `scope="col"` for WCAG 2.2 SC 1.3.1 (Info and Relationships). Affects all weekday header cells.

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -0,0 +1,366 @@
...
+                            rangeStart && a11yRangeText?.start,
+                            inRange && a11yRangeText?.in,
+                            rangeEnd && a11yRangeText?.end,
+                          ]
+                            .filter(Boolean)
+                            .join(" - ")
+                        )
+                        tabindex=(focus !== iso && -1)
+                        aria-current=iso === today && "date"
+                        aria-pressed=isSelected && "true"
```
ArtBlue, line 269: **Non-blocking:** The ARIA APG date-picker grid pattern recommends putting selection state on the cell (`<td role="gridcell" aria-selected=isSelected>`) rather than using `aria-pressed` on the button inside it. `aria-pressed` is the toggle-button pattern and carries different semantics. Consider moving to `aria-selected` on the `<td>` if you go with `role="grid"` (see blocking comment above).

packages/evo-marko/src/tags/evo-calendar/index.marko
LuLaValva: We should talk about this separately, ebayui-core uses table semantics so we shouldn't be hasty about making breaking SR changes in the new version.

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -0,0 +1,395 @@
...
+  /**
+   * Must be present to allow navigation between months
+   *
+   * Should return a string like `"Show ${month}"`
+   */
+  a11yNavigateText?: (monthName: string, dir: "next" | "prev") => string;
+  /**
+   * Clipped text to communicate range cells to screen readers, required if there is or will be a selected range
+   */
+  a11yRangeText?: {
```
HenriqueLimas, line 35: Maybe `a11yRangeTexts` (with the `s`) so it is explanatory that is multiple ones

```suggestion
  a11yRangeTexts?: {
```

HenriqueLimas (approved):

packages/evo-marko/src/tags/evo-calendar/index.marko
LuLaValva: It's not multiple though! It's presented like `10 - in range`

HenriqueLimas: `startText`, `endText`, `inText` are multiple 😄

PR #683, opened 2026-05-18
feat(badge): support empty, update styles to match specs

### LINKED ISSUE #498: Badge enhancements
### Description

Update to maintain the circle shape for single and double digit numbers
Shift to squircle for three digits plus
introduce the single dot as a variant (tiny dot with no number)

**Action**
Test zoom within browser to maintain styling

[Figma link](https://www.figma.com/design/eXvuGFnJoonyN0xN351dG2/Badge-update?node-id=15058-0)

### Screenshots

_No response_

### Figma link

https://www.figma.com/design/eXvuGFnJoonyN0xN351dG2/Badge-update?node-id=15058-0

- [ ] Supports dark mode.

LuLaValva:
- Fixes #498 

## Description

- Small badge when `:empty`
- Must use `.badge--wide` for badges that have more than 2 digits

## Notes

- I moved to a more verbose code style in Marko & React. It's a little messier, but I _mostly_ chose to do it this way because I'm afraid of hydration/comment markers being generated inside the `<span>` which would interfere with `:empty`
- `ebayui-core` and `ebayui-core-react` have some code to round the number to an integer. I think for `evo-marko` and `evo-react` we can trust that an integer is passed in

## Screenshots

image

## Checklist

- [x] I verify the linked issue has been triaged ("Needs Triage" label removed)

- [x] I tested the UI in dark mode and RTL mode


packages/evo-marko/src/tags/evo-details/test/__snapshots__/test.server.ts.snap
```diff
@@ -211,3 +211,67 @@ exports[`details > renders small version 1`] = `
   </div>
 </details>"
 `;
+
```
LuLaValva, line 214: Looks like this file was carried over because I didn't update it in a previous PR 😓, I think it's okay to keep in as a bonus change. I'll look into why CI didn't catch this last time

packages/ebayui-core/src/components/ebay-badge/index.marko
```diff
@@ -14,14 +14,18 @@ $ const {
...
 </if>
+<else-if(+number > 99)>
+    <span ...processHtmlAttributes(htmlInput) role=role class=["badge", "badge--wide", inputClass]>
+        99+
+    </span>
+</else-if>
+<else-if(+number > 0)>
+    <span ...processHtmlAttributes(htmlInput) role=role class=["badge", inputClass]>
+        ${number}
+    </span>
```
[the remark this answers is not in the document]
LuLaValva: For `evo-marko` and `evo-react` I think we are going to trust that an integer is passed in

packages/skin/src/sass/badge/badge.scss
```diff
@@ -7,10 +7,19 @@
...
     white-space: nowrap;
+    width: 2.2em;
+}
+
+.badge:empty {
+    height: 1em;
+    width: 1em;
+}
+
+.badge--wide {
```
HenriqueLimas: I don't think we should have this class. This forces us to have JS logic to being able to style the content based on the number size passed. That should be something that CSS should handle automatically. Now, the only issue with `:empty` is the white space, comments are not considered by the `:empty`. If the framework adds white spaces, I think its an issue on the framework side that is not allowing the usage of the `:empty` css selector in a span.

packages/skin/src/sass/badge/badge.scss
```diff
@@ -7,10 +7,19 @@
     border-radius: 16px;
     color: var(--color-foreground-on-attention);
     display: inline-flex;
-    font-size: 10px;
-    height: 16px;
+    font-size: var(--font-size-smallest);
+    height: 2.2em;
```
HenriqueLimas: I think we don't use relative sizes. @ianmcburnie any context on that?

packages/skin/src/sass/badge/badge.scss
LuLaValva: ### Why I used `--wide`

I don't think CSS _can_ handle this automatically and exactly match the design spec— DS is hoping for the badge to be **exactly circular** until the content shows `99+` or a different 3-digit number.

The old badge has a `min-width` which means _narrow_ numbers are circular, but wider ones become oblong:
image
One thing we can do to make this _less_ of a breaking change is to have this current behavior as the default, and use `--circular` for 2-digit numbers. I think this is a good idea, and I'll change it to this.

### Dealing with whitespace & placeholders for `:empty`

This is why I used an `if`/`else` for `<span/>`/`<span>${number}</>` instead of `<span>${number ? number : undefined}</>. I understand it's brittle, maybe we can use a class instead. Would like to hear thoughts on this

packages/skin/src/sass/badge/badge.scss
LuLaValva: Which semantic token would you use for height? `spacing` sort of makes sense, but it uses `px` while `font-size` uses `rem` so I fear it would break visually under certain zoom scenarios

packages/skin/src/sass/badge/badge.scss
LuLaValva: Okay, I updated to use `--circle` instead of `--wide` for better backwards compat but we _do_ still need to have runtime that adds an extra class if we want to match DS

HenriqueLimas (dismissed):

saiponnada (dismissed):
LGTM!

saiponnada (approved):

PR #688, opened 2026-05-19
feat(evo-react): migrate ebay-ccd to evo-ccd

HenriqueLimas:
- Fixes #

## Description

Migrates `ebay-ccd` from `@ebay/ebayui-core-react` to `@evo-web/react` as `EvoCCD`, and adds `units` prop support to the existing `evo-ccd` Marko component.

**`@evo-web/react` — new `ccd` component:**
- `EvoCCD` component with `types.ts`, `index.ts`, `README.md`, and co-located `ccd.stories.tsx`
- Skin CSS imported directly (`@ebay/skin/ccd.mjs`); icon skin is transitive via `EvoIcon`
- `a11yText: string | null` is now required (was optional with auto-computed fallback)
- `secondaryText` removed — secondary label hardcoded to `"USB PD"` matching evo-marko
- `a11yUnits` removed — caller owns the full accessible label via `a11yText`
- `secondaryType="none"` sentinel removed — omit the prop instead
- Browser tests (`vitest-browser-react`) and SSR snapshot tests added

**`@evo-web/marko` — `evo-ccd` units prop:**
- Added `units?: string` (default `"W"`) to the `Input` interface
- Replaced hardcoded `<div>W</div>` with `<div>${units}</div>`
- Story and server snapshot test updated

**Skills updated:**
- `evo-migrate-react`: documented that `undefined` must not appear in `select` control options
- `evo-migrate-react`: documented that `@ebay/skin/icon.mjs` must not be imported directly — transitive via `EvoIcon`
- `evo-app-migrate-react`: added `evo-ccd.md` migration guide and index row

## Notes

N/A

## Screenshots

N/A — no visual changes

## Checklist


packages/evo-react/src/ccd/ccd.tsx
```diff
@@ -0,0 +1,45 @@
...
+    units = "W",
+    secondaryType,
+    a11yText,
+    className,
+    ...rest
+}: EvoCCDProps) {
+    return (
+        <div
+            {...rest}
+            role="figure"
```
LuLaValva: We recently needed to change `figure` to `image` for the role, can you make sure this is up to date?

- #692

packages/evo-react/src/ccd/ccd.tsx
```diff
@@ -0,0 +1,45 @@
...
+            aria-label={a11yText ?? undefined}
+            className={classNames("ccd", className)}
+        >
+            {chargerIcon === "included" && (
+                <EvoIconCcdChargerIncluded className="ccd__charger-icon" />
+            )}
+            {chargerIcon === "not-included" && (
+                <EvoIconCcdChargerNotIncluded className="ccd__charger-icon" />
+            )}
+            {(min || max) && (
```
LuLaValva, line 31: Was just chatting with @saiponnada , it seems dangerous to have a CCD which has _only one_ of min or max since ` - X Watts.` and `X -   Watts.` are both invalid. What do you think about changing the API from
```marko
<evo-ccd min=10 max=20/>
```
to
```marko
<evo-ccd range=[10, 20]/>
```

I think this would be easier to enforce with TS and also less prone to errors

LuLaValva (commented):
Looks great! One mandatory update and one recommended API change, what do you think?

packages/evo-react/src/ccd/ccd.tsx
HenriqueLimas: It seems that min and max, when present they need to be used together so it should be a && condition and a type that if one is passed the other also needs to be passed. I do think we should maintain `min`, `max` since that is the terms used in the spec too

https://eur-lex.europa.eu/eli/dir/2022/2380/oj/eng

LuLaValva: Okay! We can keep min/max but we should switch to `&&`. We also _could_ make it type safe with a union type, but that might be overkill

LuLaValva (approved):

PR #706, opened 2026-06-08
260301 - Icon removals

### LINKED ISSUE #544: 260301 - Icon removals
### Description

We have an icon that has been renamed, so we need to remove:

- Bank group logo 24 _-colored_ 
- This icon has been replaced with `Pay by bank uk 24 -colored`

### Screenshots

_No response_

### Figma link

_No response_

- [x] Supports dark mode.
- [x] Supports responsive design.
- [x] Tokens have been provided (or already exist).

ArtBlue:
Fixes #544 — adds pay-by-bank-uk-24-colored, deprecates bank-group-logo-24-colored.


src/data/icons.json
```diff
@@ -3445,9 +3445,17 @@
...
+      "bank-group-logo-24-colored"
+    ],
+    "deprecated": [
+      {
+        "name": "bank-group-logo-24-colored",
+        "size": "24-colored",
+        "height": "24",
+        "width": "83"
+      }
     ],
```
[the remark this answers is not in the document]
ArtBlue: This is no accurate. The script was the one that was not taking the array of objects into consideration. I'll fix that. The array of objects is necessary in order to display the deprecated icon in the deprecated section.

package.json
```diff
@@ -20,6 +20,7 @@
     "lint": "stylelint src --config .stylelintrc",
     "prepare": "husky",
     "release": "npm run release --ws --if-present && changeset publish",
+    "site": "\"marko-run\"",
```
LuLaValva: This doesn't really belong here, but if we do want to add a way to run the site without `concurrently` as a bonus change we should use the Marko Run conventions

```suggestion
    "dev": "marko-run dev",
```

ArtBlue: Taking a broader approach around scripts.

LuLaValva (commented):
It looks like this isn't using the built-in scripts for updating SVG icons in each component library. Marko libraries have `importSVG` and React ones have `update-icons`. When I run these scripts locally I see file updates across the board.

package.json
```diff
@@ -20,8 +20,8 @@
     "lint": "stylelint src --config .stylelintrc",
     "prepare": "husky",
     "release": "npm run release --ws --if-present && changeset publish",
-    "site": "\"marko-run\"",
-    "start": "concurrently --kill-others-on-fail \"marko-run\" \"npm start -w packages/ebayui-core\" \"npm start -w packages/ebayui-core-react\"",
+    "start": "marko-run dev",
+    "start:all": "concurrently --kill-others-on-fail \"marko-run dev\"  \"npm run storybook -w packages/skin\" \"npm run storybook -w packages/ebayui-core\" \"npm run storybook -w packages/ebayui-core-react\"",
```
LuLaValva, line 24: This won't work as-is, since `npm run storybook` doesn't exist in our component libraries (they use `npm run start`). I think `storybook` is better but we would need to update those 4 `package.json` to make it work

ArtBlue: Did you try it? It worked for me. I can revert it if not.

LuLaValva (approved):

PR #711, opened 2026-06-11
refactor(evo-marko): make apis more consistent

LuLaValva:
Since `evo-marko` is still in beta, now is the time to make breaking changes to ensure that APIs are consistent across the board. As an experiment, I ran an `autoresearch` agent overnight and had it look for any names that don't match between components and it did a pretty good job!

### DIFF
--- a/packages/evo-marko/src/tags/evo-input/index.marko
+++ b/packages/evo-marko/src/tags/evo-input/index.marko
@@ -43,7 +43,7 @@ export interface Input extends Marko.HTML.Input {
 
 <evo-floating-label|{ showPlaceholder }|
     floatingLabel=floatingLabel
-    inputSize=size
+    size=inputSize
     alwaysUp=floatingLabelStatic
     htmlFor=inputId
     disabled=disabled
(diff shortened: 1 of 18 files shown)


packages/evo-marko/src/tags/evo-input/index.marko
```diff
@@ -21,7 +21,7 @@ export interface Input extends Marko.HTML.Input {
     floatingLabelStatic,
     fluid,
     id,
-    inputSize,
+    size,
```
HenriqueLimas: This will overwrite the `size` input attribute, maybe `scale` instead.

packages/evo-marko/src/tags/evo-textarea/index.marko
```diff
@@ -31,7 +31,7 @@ export interface Input extends Marko.HTML.TextArea {
 
 <evo-floating-label|{ showPlaceholder }|
     floatingLabel=floatingLabel
-    inputSize=inputSize
+    inputSize=size
```
HenriqueLimas: instead `size=size`?

HenriqueLimas: this needs to be reverted too

```suggestion
    inputSize=inputSize
```

Types should fail here 🤔

saiponnada (approved):

HenriqueLimas (approved):

PR #716, opened 2026-06-15
ebay-filter-menu-button: aria-pressed incorrectly rendered on menu item selection

### LINKED ISSUE #693: ebay-filter-menu-button: aria-pressed incorrectly rendered on menu item selection
### I verified there's no existing issue for this bug.

- [x] There are no existing issues

### Current behavior

When a menu item is selected in ebay-filter-menu-button, the button element renders an aria-pressed state attribute. This is incorrect behavior for this component type

### Expected behavior

The aria-pressed attribute should not be present on the ebay-filter-menu-button element when a menu item is selected. The aria-pressed attribute is intended for toggle buttons

### Package and version

Latest

### Device/OS/Browser info

N/A (DOM/markup issue, not device-specific) 

### Steps to reproduce

```bash
1. Render an ebay-filter-menu-button component with menu items 
2. Select a menu item
3. Inspect the button element's aria attributes in the DOM
4. Observe that aria-pressed is incorrectly set on the button element
```

### Relevant code block

_No response_

### Additional context

Image

ArtBlue:
Fixes #693

---
*Generated by Local Conductor · QA-validated before push*


LuLaValva:
This isn't a complete solution to the problem. Looks like we already _do_ have a complete solution in skin ([see here](https://opensource.ebay.com/evo-web/components/filter-chip/css#menu-button-chip)), but we haven't yet updated Marko or React components.

Here is what actually needs to happen:

1. Use `filter-chip--selected` instead of creating a new class
1. Add `<span class="filter-chip__count">(+x)</span>` and/or a way of letting devs modify `__count` themselves
1. Require i18n a11y text like "Filter Applied" and put it in `<span class="clipped">`

ArtBlue:
> This isn't a complete solution to the problem. Looks like we already _do_ have a complete solution in skin ([see here](https://opensource.ebay.com/evo-web/components/filter-chip/css#menu-button-chip)), but we haven't yet updated Marko or React components.
> 
> Here is what actually needs to happen:
> 
> 1. Use `filter-chip--selected` instead of creating a new class
> 2. Add `<span class="filter-chip__count">(+x)</span>` and/or a way of letting devs modify `__count` themselves
> 3. Require i18n a11y text like "Filter Applied" and put it in `<span class="clipped">`

Great catch!

packages/ebayui-core/src/components/ebay-filter-menu-button/index.marko
```diff
@@ -3,8 +3,10 @@ $ const {
     variant,
     class: inputClass,
     a11yText,
+    a11yFilterAppliedText = "Filter Applied",
     a11yFooterText,
     text,
+    count,
```
LuLaValva: A few things:

1. I don't see types for either of these new inputs, are you sure it passes CI?
2. If `hasSelected` is derived from the checked values, `count` should be too
    - If we think there are circumstances where `count` _can't_ be derived, then there are certainly also cases where `hasSelected` can't be
    - We need make this opt-in if we want to prevent breaking changes (I assume some users have already added `(+n)` to their `input.text`)

packages/skin/src/sass/filter-menu-button/filter-menu-button.scss
```diff
@@ -54,7 +54,7 @@ button.filter-menu-button__button {
     transform: rotate(180deg);
 }
 
-button.filter-menu-button__button[aria-pressed="true"] {
+button.filter-menu-button__button--active {
```
LuLaValva: We should probably call this `--selected` to align with `filter-chip--selected`

packages/ebayui-core/src/components/ebay-filter-menu-button/index.marko
```diff
@@ -29,19 +32,24 @@ $ const {
     <button
         key="button"
         type="button"
-        class="filter-menu-button__button"
+        class=["filter-menu-button__button", "filter-chip", hasSelected && "filter-menu-button__button--active", hasSelected && "filter-chip--selected"]
```
LuLaValva: Why are we adding `filter-chip` and `filter-chip--selected`? I don't think adding `chip` styles is in-scope for this issue

LuLaValva:
@ArtBlue I made a mistake, it was _not_ a great catch! We had a solution for selected `chip` but not for `filter-menu-button`. We shouldn't be adding `chip` styles to this component

LuLaValva:
The proper fix here had a few out-of-scope changes, but in the interest of doing things right I think they're worth keeping. Here they are from least → most controversial

- Since `aria-selected` isn't valid, we needed to add a `a11y-filter-applied-text=` attribute to communicate the selected state to SR
- An alternative method for communicating selected state is with "count text", which was present in evo/skin docs but not in this component. This can be provided via the new `count-text=` attribute
- The selected state was previously _derived_, which works in most cases but not when there is filtering or other custom behavior. The new `selected=` attribute allows users to override the derived selected state.

HenriqueLimas (approved):

PR #724, opened 2026-06-18
feat(evo-react): add evo-calendar

HenriqueLimas:
- Fixes #

## Description

- Add `EvoCalendar` to `@evo-web/react` with static, day selection, and range selection modes.
- Align evo-react/evo-marko calendar navigation behavior with the existing React calendar/browser datepicker-style navigation.
- Add grouped disable support, React-style controlled/uncontrolled calendar state, custom static day links via `dayLinkAs`, Storybook docs, tests, README, and changesets.
- Update app migration docs for `ebay-calendar` and package-name references.

## Notes

- No `--no-verify` bypasses were used.
- Root build completed during the push hook.

## Screenshots

N/A — no screenshots captured.

## Checklist


packages/evo-react/src/calendar/test/test.browser.tsx
```diff
@@ -0,0 +1,263 @@
...
+    it("renders linkable days with dayLinkAs when provided", async () => {
+      function DayLink({ iso, ...rest }: DayLinkAsProps) {
+        return <a {...rest} data-custom-link="true" href={`/custom/${iso}`} />;
+      }
+
+      const screen = await render(
+        <EvoCalendar
+          today="2025-01-15"
+          linkBuilder={(iso) => (iso === "2025-01-15" ? `/day/${iso}` : false)}
+          dayLinkAs={DayLink}
```
HenriqueLimas, line 258: We don’t need linkBuilder when as is passed

packages/evo-react/src/calendar/calendar.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+    onViewStartChange: {
+      action: "onViewStartChange",
+      description:
+        "Triggered when month navigation changes the first displayed month.",
+      table: { category: "Events" },
+    },
+  },
+  args: {
+    selectMode: "day",
+    today: "2025-01-15",
```
HenriqueLimas: We should use current date instead

packages/evo-react/src/calendar/calendar.tsx
```diff
@@ -0,0 +1,619 @@
...
+            >
+              <EvoIconChevronRight24 />
+            </EvoIconButton>
+          </div>
+        </div>
+      )}
+      <div
+        className="calendar__body"
+        onFocus={() => {
+          setBodyHasFocus(true);
```
HenriqueLimas, line 378: I think we can use a ref instead

packages/evo-marko/src/tags/evo-calendar/index.marko
```diff
@@ -387,6 +424,94 @@ static function isDisabled(
...
+  return today.slice(0, 7) >= viewStart && today.slice(0, 7) <= lastVisibleMonth
+    ? today
+    : ((viewStart + "-01") as DayISO);
+}
+static function offsetMonthForDay(iso: DayISO, offset: -1 | 1) {
+  const targetMonth = monthOffset(iso.slice(0, 7) as MonthISO, offset);
+  const { year, month, length } = getMonthInfo(targetMonth, 0);
+  const day = Math.min(Number(iso.slice(8, 10)), length);
+  return toISO(new Date(Date.UTC(year, month, day)));
+}
```
LuLaValva: There are a _lot_ of utilities in here, do you think we can move some that seem less component-specific to `date-utils.ts`?

It would also be great if we could find a way to share these utilities between React and Marko, I know there isn't a clean way of doing that right now 😓

packages/evo-react/src/calendar/calendar.tsx
HenriqueLimas: No we can't

packages/evo-react/src/calendar/calendar.stories.tsx
LuLaValva: If any stories are used for snapshots, we should keep a fixed date. I don't _think_ this one is so we should be in the clear, but please keep that in mind!

packages/evo-marko/src/tags/evo-calendar/date-utils.ts
LuLaValva, line 1: We already have date utils, and it's already modified in this PR! Look at `common/dates`

HenriqueLimas: These were more related to the tag it self, so it has event handlers, visible and focus handling. The other are more date related. I can move them there too.

HenriqueLimas: Merging for now, we can move it if you think its better

PR #758, opened 2026-07-16
feat(character-count): add Evo React component

HenriqueLimas:
- Fixes #

## Description

- add `EvoCharacterCount` to `@evo-web/react` with grapheme-aware counting, custom content, and input `aria-live` management
- align the experimental Evo Marko API around `text`, numeric count overrides, tag parameters, and native input references
- add React and Marko stories, browser tests, SSR tests, migration guidance, and package changesets

## Notes

- Forwarding native input references through `evo-input` and `evo-textarea` is intentionally out of scope pending guidance from the Marko team.

## Screenshots

N/A — no CSS visual changes.

## Checklist

### DIFF
--- a/packages/evo-marko/src/tags/evo-character-count/index.marko
+++ b/packages/evo-marko/src/tags/evo-character-count/index.marko
@@ -42,20 +42,20 @@ static function countGraphemes(value: string): number {
   const inputElement = inputRef?.();
   if (inputElement) {
     const previousAriaLive = inputElement.getAttribute("aria-live");
+    $signal.onabort = () => {
+      if (previousAriaLive === null) {
+        inputElement.removeAttribute("aria-live");
+      } else {
+        inputElement.setAttribute("aria-live", previousAriaLive);
+      }
+    };
+  }
+</script>
+
+<script>
+  const inputElement = inputRef?.();
+  if (inputElement) {
     inputElement.setAttribute("aria-live", ariaLive);
-    $signal.addEventListener(
-      "abort",
-      () => {
-        if (previousAriaLive === null) {
-          inputElement.removeAttribute("aria-live");
-        } else {
-          inputElement.setAttribute("aria-live", previousAriaLive);
-        }
-      },
-      {
-        once: true,
-      },
-    );
   }
 </script>
 
--- a/packages/evo-marko/src/tags/evo-character-count/index.marko
+++ b/packages/evo-marko/src/tags/evo-character-count/index.marko
@@ -1,44 +1,75 @@
 export interface Input extends Omit<Marko.HTML.Span, "content"> {
-    string: string;
-    count?: number;
-    max: number;
-    /**
-     * Clipped text for screen readers. Often something like `"characters used"`
-     *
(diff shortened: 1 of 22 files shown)


packages/evo-marko/src/tags/evo-character-count/index.marko
```diff
@@ -1,44 +1,75 @@
...
-            </span>
-        </if>
-    </span>
-</define>
+<script>
+  const inputElement = inputRef?.();
+  if (inputElement) {
+    const previousAriaLive = inputElement.getAttribute("aria-live");
+    inputElement.setAttribute("aria-live", ariaLive);
+    $signal.addEventListener(
```
LuLaValva: Marko effect convention: Use `signal.onabort = …` instead of `addEventListener("abort", …)`

Also, this doesn't actually work like I think you are hoping. If there was originally no `aria-live` and `ariaLive` was toggled at least once before this element was cleaned up, then it will just revert to the previous toggle.

packages/evo-react/src/character-count/character-count.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const count = inputCount ?? countCharacters(text ?? "");
+  const isOverLimit = count > max;
+
+  useEffect(() => {
+    const input = inputRef?.current;
+    if (!input) {
+      return;
+    }
+
+    const previousAriaLive = input.getAttribute("aria-live");
```
LuLaValva, line 24: This has the same `previousAriaLive` bug as Marko, if it's toggled at least once.

packages/evo-marko/src/tags/evo-character-count/index.marko
```diff
@@ -1,44 +1,75 @@
...
 }
 static const segmenter = (
-    "Segmenter" in Intl
-        ? new (Intl as any).Segmenter(undefined, { granularity: "grapheme" })
-        : null
+  "Segmenter" in Intl
+    ? new (Intl as any).Segmenter(undefined, { granularity: "grapheme" })
+    : null
 );
 static function countGraphemes(value: string): number {
```
LuLaValva, line 19: Can we export this, since we did the same in React?

We should also name them consistently, either both `countGraphemes` or both `countCharacters`. I lean towards `countGraphemes` because it's clear there's something more than `str.length` happening

HenriqueLimas: I think we don't need to export in Marko as we have the tag param already. I exported in react as I want to avoid render props as much as possible.

packages/evo-marko/src/tags/evo-character-count/index.marko
```diff
@@ -1,44 +1,75 @@
...
-        ${count}/${max}
-        <if=a11yText>
-            ${" "}
-            <span class="clipped">
-                ${a11yText}
-            </span>
-        </if>
-    </span>
-</define>
+<script>
```
LuLaValva: Script execution order based on source order is actually _not_ a promise in Marko. This is one of the few places besides external libraries where I actually think `<lifecycle>` makes more sense.

LuLaValva (approved):

PR #762, opened 2026-07-17
ci: parallelize CI into per-package jobs with dependency caching

LuLaValva:
In theory, this should speed up the CI by a significant amount by caching initial downloads and parallelizing other tasks. Let's test it out!

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,43 @@
+name: Setup dependencies
+description: >
+  Restore node_modules from cache (skipping npm ci entirely on a hit) and
+  optionally install the Playwright chromium browser for jobs that run
+  browser tests.
+inputs:
+  playwright:
+    description: Install the Playwright chromium browser ("true"/"false")
+    required: false
+    default: "false"
+runs:
+  using: composite
+  steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v6
+      with:
+        node-version: 24
+        cache: npm
+    - name: Restore node_modules
+      id: node-modules
+      uses: actions/cache@v4
+      with:
+        path: |
+          node_modules
+          packages/*/node_modules
+        key: node-modules-${{ runner.os }}-node24-${{ hashFiles('package-lock.json') }}
+    - name: Restore Playwright browsers
+      if: inputs.playwright == 'true'
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/ms-playwright
+        key: playwright-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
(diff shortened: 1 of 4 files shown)


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,43 @@
...
+  using: composite
+  steps:
+    - name: Setup Node.js
+      uses: actions/setup-node@v6
+      with:
+        node-version: 24
+        cache: npm
+    - name: Restore node_modules
+      id: node-modules
+      uses: actions/cache@v4
```
ArtBlue, line 21: **Non-blocking:** Third-party actions (`actions/cache@v4`, `actions/setup-node@v6` above) are pinned by mutable major-version tag rather than a full commit SHA. Since this composite action is now the shared dependency-setup path for every CI job, consider pinning to a SHA (with a `# vX.Y.Z` comment) for supply-chain hardening — a tag can be moved to point at different code without your review.

.github/workflows/preview.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
-      deployments: read
+      pages: read
```
ArtBlue, line 19: **Non-blocking:** This PR's description only mentions caching/parallelization, but this file's "wait for deployment" mechanism is being replaced wholesale (Deployments-API polling → polling a stamp file served from the site itself) along with dynamic Pages URL resolution. That's a good fix (the API apparently doesn't reliably record these deploys), but it'd help reviewers to mention it explicitly in the PR description rather than have it show up only as an incidental diff.

ArtBlue (commented):
Well-engineered refactor overall: the per-package job split, shared `setup` composite action, and node_modules/Playwright caching all look correct, and the concurrent-cache-key collision case (multiple jobs racing to write the same cache key) is a benign no-op per `actions/cache`'s own semantics. The `preview.yml` rewrite (dynamic Pages URL resolution + stamp-file-based deploy verification, replacing the old Deployments-API polling) is solid, well-commented engineering, but it's a materially bigger change than the PR description implies ("caching" + "parallelizing") — worth calling out explicitly in the description so reviewers know they're also reviewing a fix to a previously flaky preview-readiness check. No blocking issues found; two non-blocking suggestions inline.

HenriqueLimas (commented):
I think this is great and make the build faster but I also think the some packages need to have skin built before building/testing them. At this point maybe using pnpm is a better and simpler approach since they will run in parallel on the right order depending on the dependency graph instead of manually defining them in the package.json or a github action.

HenriqueLimas:
We are going to move to pnpm instead of this so it does automatically builds the correct dependency graph between packages

PR #776, opened 2026-07-28
feat(site): restore theme switcher for Evo/Evo Live

### LINKED ISSUE #624: Add theme switcher to Evo Web site
Prior to the new site rebrand, we had a theme switcher in the topbar. 

For this issue, add that back. 

The switcher should be visible and usable on all screens - small and large.

ArtBlue:
## Summary
- Restores the Evo / Evo Live theme switcher that was removed during the site rebrand (Fixes #624)
- Implemented as vanilla HTML/CSS reusing the site's existing `listbox-button` widget infrastructure (`main.js`'s legacy MakeupJS wiring), rather than the `<evo-listbox-button>` Marko component — that component's source lives outside the root `tsconfig.json`'s `rootDir`, which caused `npm run build`'s type-check step to corrupt unrelated `evo-marko` source files
- Selection persists via `localStorage`, with an inline `<html-script>` in `+layout.marko` applied before paint to avoid a flash of the wrong theme on load
- Two instances render: desktop nav (`.theme-switcher--nav`, hidden below 768px) and the mobile off-canvas panel (`.mobile-theme-switcher`)

## Test plan
- [x] Verified in-browser (desktop + mobile off-canvas instances): opens on first click, closes on selection, updates button text, toggles `evo-theme-live` on `<body>`
- [x] Verified `localStorage` persistence and correct restored state (button text, `aria-selected`, body class) across a reload with no flash of the wrong theme
- [x] Verified bidirectional toggling (Evo ↔ Evo Live) in both instances
- [x] Verified zero console errors throughout
- [x] `npm run build` passes clean (no cross-package rootDir issues)

### DIFF
--- a/src/tags/site-header/tags/theme-switcher/index.marko
+++ b/src/tags/site-header/tags/theme-switcher/index.marko
@@ -1,0 +1,84 @@
+import "./style.scss";
+
+export interface Input extends Marko.HTML.Div {}
+
+<const/{ class: inputClass, ...htmlInput }=input>
+
+<div ...htmlInput class=["theme-switcher", inputClass]>
+    <span class="listbox-button listbox-button--form">
+        <button
+            class="btn btn--form"
+            type="button"
+            aria-expanded="false"
+            aria-haspopup="listbox">
+            <span class="btn__cell">
+                <span class="btn__label">
+                    Theme:
+                </span>
+                <span class="btn__text">
+                    Evo
+                </span>
+                <svg class="icon icon--12" height="10" width="14" aria-hidden="true">
+                    <icon-symbol name="chevron-down-16"/>
+                </svg>
+            </span>
+        </button>
+        <div class="listbox-button__listbox">
+            <div class="listbox-button__options" role="listbox">
+                <div
+                    class="listbox-button__option"
+                    role="option"
+                    data-value="evo"
+                    aria-selected="true">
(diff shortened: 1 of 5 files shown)


src/tags/site-header/tags/theme-switcher/index.marko
```diff
@@ -0,0 +1,84 @@
+import "./style.scss";
+
+export interface Input extends Marko.HTML.Div {}
+
+<const/{ class: inputClass, ...htmlInput }=input>
+
+<div ...htmlInput class=["theme-switcher", inputClass]>
+    <span class="listbox-button listbox-button--form">
+        <button
+            class="btn btn--form"
```
HenriqueLimas, line 10: I think that a `btn--borderless` looks cleaner


Image


```suggestion
    <span class="listbox-button">
        <button
            class="btn btn--borderless"
```

ArtBlue: @HenriqueLimas , I was deciding between the two, and went with the default version so it would be clear affordance of a selection. The borderless relies a bit too much on the caret IMO. I do agree it is cleaner though.

HenriqueLimas: Make sense, I think this is just taste, so its relative both are fine to me

HenriqueLimas (approved):

PR #778, opened 2026-07-29
docs(storybook): align Evo branding

HenriqueLimas:
- Fixes #

## Description

- Replace the eBay logo in Evo Marko and Evo React Storybook with the Evo Web icon and wordmark.
- Align both Storybook manager palettes with the current Skin light-theme accent and secondary colors.

## Notes

- Storybook-only configuration change; no package changeset is required.
- Verified with both targeted Storybook builds. The repository pre-push hook also completed successfully.

## Screenshots

Screenshot 2026-07-28 at 6 00 27 PM

## Checklist


ArtBlue (dismissed):
This is sweet!!!! Evoooooo....

HenriqueLimas:
Also adding dark/light mode support for evo-react but moving to a private evo-storybook-addon

packages/evo-storybook-addon-theme/src/theme.ts
```diff
@@ -0,0 +1,58 @@
...
+  fontBase: '"Market Sans", Arial',
+  brandTitle: "Evo Web",
+  brandUrl: "/",
+  brandTarget: "/",
+};
+
+export const lightTheme = create({
+  ...brand,
+  base: "light",
+  colorSecondary: "#0968f6",
```
ArtBlue: **Non-blocking:** `#0968f6`/`#4d93fc` (line 39) match Skin's `--color-blue-500`/`--color-blue-400` core tokens, but that's not documented here. Since Storybook's `theme.ts` can't consume CSS custom properties directly, hardcoding is reasonable, but a short comment naming the source tokens would help whoever updates this the next time the Skin accent color changes — right now there's nothing tying these hex values back to the design system.

packages/evo-storybook-addon-theme/src/manager.tsx
```diff
@@ -36,17 +66,25 @@ new MutationObserver(applyToIframe).observe(document.documentElement, {
...
 
-  // Sync persisted globals → module variable on every render.
-  // Skip undefined during view-mode transitions to avoid clobbering
-  // an active selection. The onReset callback handles intentional resets.
-  if (current !== undefined) {
-    currentScheme = current;
-  }
-  applyToIframe();
+  // Keep the manager and preview synchronized with Storybook's persisted global.
+  currentScheme = current;
```
ArtBlue, line 86: **Non-blocking:** `currentScheme = current; applyColorScheme();` runs as a side effect directly in the component body rather than in a `useEffect`. It works today since Storybook's manager doesn't use React concurrent rendering, but it's fragile — consider `useEffect(() => { currentScheme = current; applyColorScheme(); }, [current]);` to avoid relying on that assumption.

ArtBlue (commented):
Nice consolidation — centralizing the Evo Web manager theme, logo, and color-scheme sync into a shared `@evo-web/storybook-addon-theme` addon (instead of duplicated per-package `theme.ts`/`manager.ts`) is the right move, and the light/dark/system sync logic (manager + preview iframe + Docs container) is thorough and well-commented. No blocking issues; two non-blocking suggestions inline, plus one coordination FYI: PR #762 (open, not yet merged) restructures CI into one job per package and doesn't currently include a job for this new `evo-storybook-addon-theme` package — worth a follow-up once that lands so this package's `build`/type-check doesn't silently stop running in CI.

packages/evo-storybook-addon-theme/src/theme.ts
HenriqueLimas: Going to use `@ebay/design-tokens/esm/evo` directly so we don't need to hard code it.

HenriqueLimas:
> FYI: PR https://github.com/eBay/evo-web/pull/762 (open, not yet merged) restructures CI into one job per package and doesn't currently include a job for this new evo-storybook-addon-theme package — worth a follow-up once that lands so this package's build/type-check doesn't silently stop running in CI.

@ArtBlue, the build of storybook addon is just a typecheck, not required for building storybook itself.

packages/evo-storybook-addon-theme/src/manager.tsx
HenriqueLimas: Not sure this is necessary, we can change in case it will have any issue

PR #781, opened 2026-07-30
feat(site): add Theming guide page (pass 1) (#625)

### LINKED ISSUE #625: Add a dedicated Theme page to the Evo Web site
## Summary

We need to provide users with more guidance on theming. A dedicated theming page seems to be the best option for this.

## Notes

Since we have a "Guides" section with guides to other helpful docs, this seems like a good parent for a "Theming Guide" page. It should probably live under `https://opensource.ebay.com/evo-web/guides/theming`.

We should also link to the theming guide page from the guides landing page: https://opensource.ebay.com/evo-web/guides

The current tokens page (https://opensource.ebay.com/evo-web/components/tokens/css) has some crossover content that should probably also move to the Theming page, specifically the Evo Live Theme and Subtheme sections.

The IA of the page should probably look a bit like this:

```
* Theming (how to create a theme)
** (sections)
* Sub-themeing (how to create a sub-theme)
** (sections)
* Current Themes
** Evo Live Theme
...
```

ArtBlue:
## Summary

Pass 1 of #625 — a dedicated `/guides/theming` page covering how to build a theme, how to sub-theme, and which themes ship today (Evo, Evo Live). Passes 2 (theme creator tool) and 3 (apply-theme/session-storage) are intentionally out of scope for this PR; see the design spec's "Explicitly Out of Scope" section.

- New `/guides/theming` page, migrating the "Evo Live Theme" and "Subthemes" content off the Tokens (CSS) page and reframing it generically
- Promoted the `token-doc` tag from a route-private folder to the shared `src/tags/` so both pages can render the semantic/live token tables
- Added a short intro blurb + "Theming Guide" button to the top of the Tokens (CSS) page, explaining what tokens are and how they relate to themes
- Linked the new guide from all 5 places this repo hand-lists guides (landing page, mobile nav, sitemap, sidebar navrail, footer) — the last two weren't caught until manual browser QA, since the original plan only accounted for 3
- Design spec and implementation plan committed under `docs/superpowers/` for reference

No changeset included — everything touched is docs (`src/routes/`) or a site nav tag (`src/tags/`), not a versioned package under `packages/`, per this repo's own changeset exemption.

## Test plan

- [x] `npm run build` passes (verified locally and via the pre-push hook)
- [x] `npx mtc` type-check passes clean
- [x] Manual browser QA (Playwright): `/guides/theming` renders both token tables, both syntax-highlighted code samples, and the sub-theme demo with visually distinct button styling, in both light and dark mode
- [x] `/components/tokens/css` confirmed to no longer show the old "Evo Live Theme"/"Subthemes" headings, with the new intro blurb + button and no orphaned content
- [x] All 5 guides-list locations (landing page, mobile nav Guides tab, sitemap, sidebar navrail, footer) confirmed linking correctly to `/guides/theming`
- [ ] Percy visual regression (CSS changes touch Tokens page + 5 nav files — reviewer to confirm Percy run on this PR)

Fixes #625 (pass 1)


src/routes/_index/components/tokens/css+page.marko
```diff
@@ -1,5 +1,24 @@
...
+    <code>--color-foreground-accent</code>. A theme is built by overriding this same set of tokens, so every
+    component picks up the new values without any component code changing.
+    The tables below are the full set of tokens available today; to learn
+    how to build your own theme or apply a subtheme, see the Theming Guide.
+  </p>
+  <p>
+    <a class="fake-btn fake-btn--secondary" href=`${urls.guides}/theming`>
+      Theming Guide
+    </a>
+  </p>
```
HenriqueLimas: I think we shouldn't use a fake-btn. Fake button are for actions like `Go to Checkout`. A normal link have a better semantic here since it is linking to another page. Better adding the link above in the `see the "<a>Theming Guide</a>`


```suggestion
    how to build your own theme or apply a subtheme, see the <a href=`${urls.guides}/theming`>Theming Guide.</a>
  </p>
```

ArtBlue: Since the new theming page -

1. Includes a lot of content the current page contained
2. Is not as discoverable from the UI
3. Is an increasingly more important page since we're getting more themes
4. Needs to be extended with additional functionality

I think it's critical that we make it a bit more discoverable. This is the reason I went with a more prominent CTA rather than a link that get lost on a page.

HenriqueLimas: TBH, I think `Guides -> Theming Guide` is way easier to discover rather than `Components -> Tokens -> CSS -> Theming Guide`.

I think that the button there doesn't look good though and its not semantically correct, we could have a section notice instead and with instruction on what to do. Just `Theming Guide` is not clear what it needs to be done, "Go to Theming Guide" is more clear. Something like:
Screenshot 2026-07-31 at 9 40 19 AM

ArtBlue: @HenriqueLimas , hmm... I'm not sure I understand your concern here. How is it not semantically correct? It's a link that takes the user to the page with the anchor of "Theming Guide" that goes to the Guides > Theming Guide page. That is as semantically correct as you can get. 😄 

This is what fake button was built for. In fact, we use it all across our own site. Countless internal devs use it in a similar way across eBay pages.

HenriqueLimas: I don’t mean in a html perspective, more on the ui perspective, seeing a button after reading the paragraph looks off (is it trying to show the theme being used in a button or should I click this button? If i click what it does?). At least we should add a “Go to theme guide” as the button text

ArtBlue: Semantics are usually dealing with the relationship of html to the meaning the tags convey to the way the user understands their usage. Semantically, our own Button component doc defines fake-btn generically, with no carve-out for "actions only":

"Finally, a link may be styled to look like a button. We call this a fake button. In this case the <a> tag is required with a valid HREF attribute."

<kbd>image</kbd>

There are numerous examples of this sort of usage across the site, but since you already conceded that you meant something else by "semantically," it's not necessary.

So, technically, there is no semantic issue here at all and a button is technically fine to use. To me, this looks to be purely an aesthetic choice. 

I think we're straining gnats here. Instead of more back-and-forth, I'll just implement your preferred solution.

PR #785, opened 2026-07-31
feat(evo-react): add dialog

HenriqueLimas:
- Fixes #

## Description

Add the generic EvoDialog component to `@evo-web/react`, aligned with the Evo Marko contract and native `<dialog>` behavior.

- Add composable banner, header, title, previous, close, main, and footer components
- Support controlled and uncontrolled state, size variants, `closedby`, closing animations, and Safari fallbacks
- Add browser and SSR coverage plus Default, Banner, Custom Banner, and Previous Storybook examples
- Document migration from `EbayLightboxDialog` and add a patch changeset

## Notes

`EvoDialog` replaces the legacy lightbox-dialog migration path. The implementation uses native React 19 refs and named subcomponent composition without child scanning.

## Screenshots

N/A — visual permutations are available in the added Storybook stories.

## Checklist


packages/evo-react/src/dialog/dialog-title.tsx
```diff
@@ -0,0 +1,28 @@
...
+  id,
+  className,
+  children,
+  ...rest
+}: EvoDialogTitleProps) {
+  const { headerId, setHeaderId } = useDialogContext();
+  const resolvedId = id ?? headerId;
+
+  if (id) {
+    setHeaderId(id);
```
ArtBlue, line 16: **Non-blocking:** `setHeaderId(id)` is called unconditionally in the render body whenever `id` is passed. This is safe only because `useControllableId`'s setter is backed by an external store (per the comment in `use-controllable-id.ts`) rather than `useState`, so it doesn't trigger React's "cannot update a component while rendering a different component" restriction. Worth a one-line comment here referencing that guarantee — without it, this looks like a call-setState-during-render bug to a future reader or reviewer unfamiliar with `use-controllable-id.ts`.

packages/evo-react/src/dialog/dialog-close.tsx
```diff
@@ -0,0 +1,34 @@
+import { useCallback } from "react";
+import classNames from "classnames";
+import { EvoIconButton } from "../icon-button";
```
ArtBlue: **Non-blocking:** `EvoIconButton` and `type NativeIconButtonProps` are both imported from `../icon-button` in separate statements (lines 3-4). Could combine into a single `import { EvoIconButton, type NativeIconButtonProps } from "../icon-button";` for consistency with the single-import style used elsewhere in this PR (e.g. `dialog-previous.tsx`).

ArtBlue (commented):
Solid, well-tested port of the evo-marko dialog contract to React 19 — native <dialog>/showModal usage, CSS-animation-driven close, Safari `closedby` fallback, and context-based header id wiring are all correctly carried over, with strong SSR + browser test coverage. No blocking issues found; a couple of small non-blocking suggestions below around documenting a subtle render-phase pattern and tightening imports.

packages/evo-react/src/dialog/dialog-title.tsx
HenriqueLimas: We do it in different places, tricky to add the comment everywhere. The future reader/ai is probably able to get that as you pointed out `useControllableId` is safe to use.

HenriqueLimas:
No major blocking. Merging.

PR #786, opened 2026-08-03
feat(evo-react): add chip

HenriqueLimas:
## Description

Adds `EvoChip` to `@evo-web/react` using a compound-component API:

```tsx
<EvoChip>
  <EvoChipText>Football</EvoChipText>
  <EvoChipDeleteButton />
</EvoChip>
```

`EvoChip` provides a `useId()`-generated text ID through context. `EvoChipText` applies that ID, and `EvoChipDeleteButton` references it with `aria-describedby`. The migration also includes native attribute/ref support, Skin and Evo icon integration, Storybook documentation, browser and SSR coverage, app migration guidance, and a patch changeset.

### API options considered

- **Keep the legacy top-level delete props:** simplest migration, but keeps delete-button concerns on the root and diverges from Evo Marko's optional delete region.
- **Use a `text` prop:** enables concise `<EvoChip text="Football" />` usage, but prevents future composition before or around the text.
- **Use a `deleteButton` JSX prop:** avoids child scanning and enforces placement, but constrains future component structure.
- **Use named sub-components (selected):** `EvoChipText` and `EvoChipDeleteButton` preserve extensibility, avoid `React.Children` scanning, and allow future content before the text. Child ordering is intentionally consumer-controlled.

## Notes

- No Skin/CSS changes; this component consumes the existing `@ebay/skin/chip.mjs` module.
- `a11yText` defaults to the English label `"delete"` and should be localized.
- Verified with `npm run build -w packages/evo-react`, the production Evo React Storybook build, and root `npm run build`.
- No GitHub issue is linked for this PR.

## Screenshots

N/A — no Skin/CSS changes; the existing chip presentation is exposed through a new React wrapper and Storybook story.

## Checklist


packages/evo-react/src/chip/chip-delete-button.tsx
```diff
@@ -0,0 +1,24 @@
...
+}: EvoChipDeleteButtonProps) {
+  const { textId } = useChipContext();
+
+  return (
+    <button
+      {...rest}
+      type="button"
+      className={classNames("chip__button", className)}
+      aria-label={a11yText}
+      aria-describedby={textId}
```
ArtBlue, line 19: **Non-blocking:** `aria-describedby={textId}` always points at the ID `EvoChipText` would set. If a consumer renders `EvoChipDeleteButton` without a sibling `EvoChipText` (the API allows any composition/order), this becomes a dangling ARIA reference. Since `React.Children` scanning was intentionally avoided, consider at least a JSDoc note on `EvoChipDeleteButton` (or `EvoChip`) that `EvoChipText` should be present for the description to resolve correctly.

packages/evo-react/src/chip/chip.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    children: {
+      control: false,
+    },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof EvoChip>;
+
+export const Default: Story = {
```
ArtBlue, line 45: **Non-blocking:** The only story shown is the chip with a delete button. Since a text-only chip (no delete button) is a first-class, tested composition (see `test.browser.tsx` "renders chip text without a delete button"), consider adding a second high-level story for that variant so the docs/Storybook surface both supported shapes.

ArtBlue (commented):
Solid addition — the compound-component API (EvoChip/EvoChipText/EvoChipDeleteButton) is well-reasoned, matches the documented a11yText convention (CLAUDE.md explicitly cites evo-chip as the `string` reference case), correctly uses native refs with no forwardRef, uses the existing chip.scss BEM classes, and has thorough browser + SSR test coverage. No blocking issues found. Two non-blocking suggestions below around the aria-describedby contract and Storybook coverage of the text-only variant.

packages/evo-react/src/chip/chip-delete-button.tsx
HenriqueLimas: A chip needs to be used with a text. Without it, its not following our design system. This is what we have in the docs `A chip that composes text and an optional delete button through named sub-components.`. It mention that the button is optional not the text though

PR #790, opened 2026-08-04
feat(evo-react): add fake menu

### LINKED ISSUE #795: evo-react: migrate ebay-fake-menu
Parent: #794
Implementation: #790

## Description

Migrate `@ebay/ui-core-react/ebay-fake-menu` to `@evo-web/react/fake-menu`. The implementation is already in review in #790; this issue provides focused tracking and should close when that PR is merged.

Use the repository's [Evo React migration guidance](https://github.com/eBay/evo-web/blob/main/.agents/skills/evo-migrate-react/SKILL.md).

## Goals

- Preserve semantic HTML, Skin/BEM styling, accessibility behavior, keyboard support, RTL behavior, and pass-through attributes.
- Align the public API with the corresponding Evo Marko contract without carrying forward obsolete React workarounds.
- Follow React 19 and `@evo-web/react` package conventions.

## Acceptance criteria

- [ ] Uses named exports, native React 19 `ref` props, and no `forwardRef`, `FC`, or default exports.
- [ ] Imports the appropriate Skin `.mjs` module directly.
- [ ] Avoids `React.Children` inspection; composition uses explicit components/context where needed.
- [ ] Public custom props and object fields have JSDoc and follow Evo accessibility naming conventions.
- [ ] Includes browser interaction tests and SSR snapshot coverage.
- [ ] Includes Storybook documentation and a minimal component README.
- [ ] Updates consumer migration guidance in `evo-app-migrate-react`.
- [ ] Includes an `@evo-web/react` changeset.
- [ ] Meets WCAG 2.2 AA and supports keyboard use and RTL.
- [ ] `npm run build` passes.

HenriqueLimas:
- Fixes #795
## Description

Adds `EvoFakeMenu` to `@evo-web/react` using the explicit compound-component API established by `EvoMenu`.

- Adds root, items, item, badge, and separator components backed by existing Skin menu styles.
- Supports anchor and button items, React 19 refs, custom framework link components through `as`, current and disabled states, class prefixes, and positioning variants.
- Updates Skin to infer badged fake-menu items with `:has(.badge)` while retaining the legacy `--badged` selectors for compatibility.
- Adds browser and SSR coverage, Storybook documentation, app migration guidance, and patch changesets for Skin and evo-react.

## Notes

No issue number was provided. No verification hooks were bypassed.

## Screenshots

Percy dry-run discovery found all eight `Skin/Menu/Fake Menu` stories, including the badged use case. No Percy build was created in dry-run mode.

## Checklist


packages/evo-react/src/fake-menu/fake-menu-item.tsx
```diff
@@ -0,0 +1,59 @@
...
+  disabled = false,
+  type,
+  href,
+  as: _as,
+  children,
+  ...rest
+}: EvoFakeMenuItemProps) {
+  const { baseClass } = useFakeMenuContext();
+  const ariaCurrent = current ? (itemMatchesUrl ? "page" : "true") : undefined;
+  const itemClassName = classNames(`${baseClass}__item`, className);
```
ArtBlue, line 20: **Blocking:** Skin's `menu.scss` requires the item element to carry a `fake-menu__item--badged` modifier class (`a.fake-menu__item--badged { position: relative; }`) so the absolutely-positioned `.badge` inside it lays out correctly — see `a.fake-menu__item--badged span.badge { position: absolute; top: 0; ... }`. This component never adds that class, and the badge (passed as an `EvoFakeMenuItemBadge` child) ends up nested inside the label `<span>` here rather than as a sibling before the tick icon, unlike both the Skin reference structure and the `evo-marko` `evo-fake-menu` tag (`<span>{content}</span><if=badgeNumber><evo-badge .../></if><evo-icon-tick-16/>`). Confirmed in this PR's own SSR snapshot: `<a class="fake-menu__item"><span>Messages<span class="badge badge--circle">5</span></span>...</a>` — no `--badged` class, badge nested in the label span. Without `position: relative` on the item, the badge's `position: absolute` will position against whatever ancestor happens to be positioned, breaking the intended overlay layout. Consider detecting an `EvoFakeMenuItemBadge` child (e.g. via `Children.toArray`) to render it as a sibling of the label span and conditionally add the modifier class, similar to how the legacy `ebayui-core-react` version derives `hasBadge` from its `badgeNumber` prop.

packages/evo-react/src/fake-menu/fake-menu-separator.tsx
```diff
@@ -0,0 +1,19 @@
+import classNames from "classnames";
+import { useFakeMenuContext } from "./context";
+import type { EvoFakeMenuSeparatorProps } from "./types";
+
+export function EvoFakeMenuSeparator({
+  className,
+  ...rest
+}: EvoFakeMenuSeparatorProps) {
```
ArtBlue, line 8: **Non-blocking:** `EvoFakeMenuSeparatorProps` omits `role` from the native `<hr>` props (implying the component controls it), but no `role="separator"` is ever set here — the `evo-marko` sibling sets it explicitly (`<hr class=... role="separator">`). Native `<hr>` does default to an implicit `separator` role in most browsers/AT, so this likely isn't a real regression today, but consider setting it explicitly to match the sibling implementation and the apparent intent behind omitting `role` from the prop type.

ArtBlue (commented):
Solid compound-component structure and thorough test/story coverage overall, but there's one blocking bug: `EvoFakeMenuItem` never applies the `fake-menu__item--badged` modifier class Skin's CSS requires for badge positioning, and nests the badge inside the label `<span>` instead of rendering it as a sibling — confirmed in the SSR snapshot (`<span>Messages<span class="badge...">5</span></span>`). The legacy `ebayui-core-react` implementation gets this right (`hasBadge && "menu__item--badged"`); this compound-API redesign lost it. Also non-blocking: the separator drops the explicit `role="separator"` its Marko sibling sets (native `<hr>` implies it, so low risk, but `EvoFakeMenuSeparatorProps` explicitly omits `role` from the props type, implying the intent was to hardcode it). Separately, both this PR's description and the migration doc say EvoFakeMenu follows "the same compound-component API as EvoMenu", but no `EvoMenu` React component exists in this repo (only `evo-marko`'s `evo-menu`, which uses a different attr-tag/slot API) — worth double-checking that reference isn't a leftover from a different draft.

packages/evo-react/src/fake-menu/fake-menu-item.tsx
HenriqueLimas: Need to make the same change as menu on skin to use `has()` instead. Same as I doing in #787

HenriqueLimas: `Children.toArray` and any `Children` method from React are deprecated and shouldn't be used.

.changeset/add-evo-fake-menu.md
```diff
@@ -0,0 +1,5 @@
+---
+"@evo-web/react": patch
+---
+
+Add EvoFakeMenu component.
```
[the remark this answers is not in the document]
HenriqueLimas: added the issue in the description

packages/evo-react/src/fake-menu/fake-menu-separator.tsx
HenriqueLimas: There is a `jsx-a11y/no-redundant-roles` in eslint that fails if we add `role="separator"`

LuLaValva (dismissed):

ianmcburnie:
There is a possibility we will be renaming this from Fake Menu (people seem to dislike and/or be confused by this name) to Navigation Menu or Disclosure Menu to align better with [APG](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/) and our forthcoming updated Playbook naming. Not sure if we want to wait, or just create an alias when needed.

HenriqueLimas:
I think that is a good idea. We might want to also review fake-tabs to NavigationTabs too. We will do in a separate PR so we rename marko/react at the same time. We don't need alias in evo-web packages as they are experimental.

PR #878, opened 2026-08-12
feat(evo-menu): replace @item separator with @group attribute tag

LuLaValva:
Adds a `<@group>` attribute tag to `evo-menu` (and passes it through `evo-menu-button`), replacing the `separator` attribute on `<@item>`. Each group takes its own `selected`/`selectedChange` and nested `<@item>` tags; separators render automatically between groups, and roving tabindex, typeahead, and arrow navigation treat the whole menu as one list. Top-level `<@item>` tags may be mixed with groups and always render first, sharing the menu-level `selected`.

`selected` (menu-level or per-group) now also accepts `null`, meaning single-select with nothing selected yet — distinct from omitting `selected`, which renders plain non-selectable menu items. Default item values are now unique across the whole menu so grouped items keep distinct `data-value`s.

Includes new `Groups` and `Controlled` stories/examples, updated tests, and a changeset (BREAKING: `separator` removed from `<@item>`).


packages/evo-marko/src/tags/evo-menu/index.marko
```diff
@@ -61,40 +100,49 @@ export interface Input<
...
-                separator,
-                badgeNumber,
-                disabled,
-                content,
-                ...htmlItem
-            }=item>
+        <for|group, groupIndex| of=groups>
+            <if=groupIndex>
+                <hr class=`${baseClass}__separator` role="separator">
+            </if>
```
HenriqueLimas, line 120: Smart 👌

HenriqueLimas (dismissed):

packages/evo-marko/src/tags/evo-menu/index.marko
```diff
@@ -61,40 +100,49 @@ export interface Input<
...
+            <let/selected:=group.selected>
+            <const/isRadio=(
+                typeof group.selected === "number" ||
+                typeof group.selected === "string" ||
+                group.selected === null
```
HenriqueLimas, line 125: Should we use `== null` instead of `=== null`.

LuLaValva: Definitely not! `null` or `string` or `number` opts this group into `radio` behavior, while `undefined` means the items should be non-selectable buttons!

PR #879, opened 2026-08-12
feat(storybook): add RTL toggle to evo-marko and evo-react storybooks

### LINKED ISSUE #14: Storybook: use RTL add-on
We should try and leverage an RTL plugin for our stories, rather than writing them ourselves.

I was looking at the one below, however it seems to lack the ability to provide a perma link to the RTL, which would prevent Percy from taking visual snapshots (Percy needs a URL). Unless there is some way for Percy to programmatically control Storybook perhaps...

https://github.com/unindented/storybook-addon-rtl

LuLaValva:
- Fixes #14

Adds an LTR/RTL text direction toggle to the shared storybook theme addon's toolbar, applied to the preview via a decorator. Defaults to System (no `dir` attribute).

### DIFF
--- a/packages/evo-storybook-addon-theme/src/preview.tsx
+++ b/packages/evo-storybook-addon-theme/src/preview.tsx
@@ -7,7 +7,29 @@ import ThemedDocsContainer from "./docs-container";
 export default {
   globalTypes: {
     colorScheme: {},
+    direction: {
+      description: "Text direction",
+      toolbar: {
+        icon: "transfer",
+        items: [
+          { type: "reset", icon: "transfer", title: "System" },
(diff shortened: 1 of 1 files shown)


HenriqueLimas:
Should we also use a RTL language to show (probably a bigger change). This is an example from shadcn:

Screenshot 2026-08-12 at 10 55 15 AM

LuLaValva:
> Should we use an RTL language

If we wanted to do that for _every_ example, then we need to implement some version of i18n 😓. For examples built _specifically_ to demonstrate RTL we should definitely use arabic or another RTL language, but I'm not sure how possible it would be to do this for all examples when RTL is enabled.

HenriqueLimas (approved):

PR #884, opened 2026-08-13
fix(evo-tooltip): open on click in browsers that do not focus buttons on click

LuLaValva:
`evo-tooltip` opens on `focusin`, but Safari and macOS Firefox do not give buttons focus on click — so a `noHover` tooltip with a button host could only be opened with the keyboard. The host element now focuses itself on click (a no-op where click already focuses), composing with any consumer-provided `onClick`. Adds the component's first browser tests, using programmatic clicks to reproduce the no-focus-on-click behavior, plus a changeset.

### DIFF
--- a/.changeset/tall-tooltips-focus.md
+++ b/.changeset/tall-tooltips-focus.md
@@ -1,0 +1,5 @@
+---
+"@evo-web/marko": patch
+---
+
+Fix `evo-tooltip` not opening on click in Safari and macOS Firefox. The tooltip opens on `focusin`, but those browsers do not give buttons focus on click, so a `noHover` tooltip with a button host could only be opened with the keyboard. The host now focuses itself on click, composing with any consumer-provided `onClick`.
--- a/packages/evo-marko/src/tags/evo-tooltip/index.marko
+++ b/packages/evo-marko/src/tags/evo-tooltip/index.marko
@@ -93,12 +93,21 @@ export interface Input extends Marko.HTML.Span {
         open = el.contains(e.relatedTarget as Node | null);
         onFocusOut && onFocusOut(e, el);
     }>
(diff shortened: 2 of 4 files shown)


HenriqueLimas (commented):
Not sure we want this behavior for clicks of buttons. Reason is that after clicking the button its too late to see the tooltip IMO.

LuLaValva:
> Not sure we want this behavior for clicks of buttons. Reason is that after clicking the button its too late to see the tooltip IMO.

I think it depends on the purpose of the button! If the result is a reversible action then it is still useful to see the button's tooltip.

This change irons out an inconsistency between browsers. _Only_ Safari doesn't trigger a focus event on button click, so the tooltip opens everywhere else. I think it's worth it to make all browsers behave the same.

HenriqueLimas:
What if Safari is the correct approach? Other libraries like [react-aria](https://react-aria.adobe.com/Tooltip) or [base-ui](https://base-ui.com/react/components/tooltip#api-reference), they both show the tooltip only on focus if hover is disabled. If we consider this sentence from [our MIND patterns](https://ebay.gitbook.io/mindpatterns/disclosure/tooltip#best-practices):

> _Tooltip should follow the principal of [progressive disclosure](http://en.wikipedia.org/wiki/Progressive_disclosure) - revealing the right information, at the right time._

Showing it on click its not at the right time IMO, its too late already and unecessary.

If we consider safari is the correct approach, we need to detect `mousedown` first and abort the tooltip to open when that happens

HenriqueLimas:
Discussing offline with @saiponnada the behavior on Safari is the expected one. So we should prevent opening of the tooltip on click of the button instead of forcing to open on click on every browser. For opening on click we have the infotip that handles that

LuLaValva:
Okay, closing since the Safari default is expected

PR #885, opened 2026-08-13
docs(evo-icon): compact all-icons storybook example into a dense grid

LuLaValva:
Replaces the evo-marko all-icons Storybook example (a spread-out grid of labeled tiles) with a dense grid: each icon is a button that toggles an `evo-tooltip` showing its selectable tag name (`<evo-icon-foo/>`), closing on focus out. The cells are generated by `importSVG` as a local `icon-example` tag with scoped styles, so nothing leaks into the global Storybook CSS and the generated example shrinks to three lines per icon. Storybook-only; no changeset.

Dense icon grid with an open tooltip showing <evo-icon-ai-camera-24/>

In Safari, closing by clicking elsewhere relies on the click-to-focus fix in #884.

### DIFF
--- a/packages/evo-marko/scripts/import-svg.ts
+++ b/packages/evo-marko/scripts/import-svg.ts
@@ -84,7 +84,10 @@ function setupDir(fileName: string) {
   cp.execSync(`rm -rf ${JSON.stringify(outputDir)}`);
   cp.execSync(`rm -rf ${JSON.stringify(example)}`);
   fs.mkdirSync(outputDir);
-  fs.writeFileSync(example, `div.icon-examples`);
+  fs.writeFileSync(
+    example,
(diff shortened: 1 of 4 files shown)


HenriqueLimas (commented):
could you add the same for evo-react 🥹

LuLaValva:
> could you add the same for evo-react 🥹

I will, but after `EvoTooltip` is merged! Otherwise we'll have to create our own pseudo-tooltip just for this case.

HenriqueLimas:
@LuLaValva I meant just the grid layout

LuLaValva:
The grid layout is great, but we need to surface the icon names somehow! I used tooltips in `@evo-web/marko`, I suppose we could do something like `title=` + copy name on click…

HenriqueLimas (dismissed):

HenriqueLimas (approved):

PR #889, opened 2026-08-14
docs: add agent-feedback system

LuLaValva:
Adds an `agent-feedback/` directory so actionable observations that are out of scope for the task that surfaced them get recorded instead of dropped — or, worse, fixed inside an unrelated diff. Ported from the marko repo, which has been running this system for a while.

Each category file (`bugs`, `a11y`, `cleanup`, `perf`, `dx`, `unclear`) starts empty; `README.md` covers when to add an entry, the entry format, and the rule that entries are deleted in the same PR as the fix. Two things differ from marko's version: an `a11y` category, since WCAG 2.2 AA is a non-negotiable here and would otherwise be buried in `bugs`, and a rule that an entry must name which of the five framework packages a defect reaches and which were actually checked.

`CLAUDE.md` gets a pointer to it, matching how marko surfaces the system.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,6 +10,7 @@ AI configuration for eBay's evo-web components monorepo.
 - Always run `npm run build` before marking component work complete
 - Prefer reading existing patterns over introducing new ones
 - Never manually edit auto-generated files (they should typically have a comment indicating they are generated)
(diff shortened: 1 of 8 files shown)


LuLaValva:
Removing in favor of a more comprehensive agent system that @ArtBlue is working on

PR #899, opened 2026-08-19
feat(card): add a generic @media attribute tag

LuLaValva:
Adds a `@media` attribute tag to the Marko `ebay-card` so consumers can put arbitrary content — a `<video>`, a canvas, a custom player — in the card's media slot instead of just an image. When present it renders in place of `@image`, which becomes optional. A `Media` story and example template cover it.

Also fixes the card image in both Marko packages: it now gets the `card__hero-image` class. Skin defines the image sizing (`height`/`width: 100%`, `object-fit: cover`), the hover transform, and the `--horizontal` / `--full-hero` overrides on that class, so the bare `` the card rendered previously picked up none of them and sat at its intrinsic size inside the media box. Neither React package has a card component, so this is Marko-only.


packages/ebayui-core/src/components/ebay-card/index.marko
```diff
@@ -52,9 +54,19 @@ $ const elem = !action ? (href ? "a" : "button") : "span";
...
     ...processHtmlAttributes(htmlInput)
 >
-    <div class="card__media">
-        <img ...processHtmlAttributes(image)>
-    </div>
+    <if(media)>
+        <div class="card__media" ...processHtmlAttributes(media)>
+            <${media}/>
+        </div>
+    </if>
```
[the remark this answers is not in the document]
LuLaValva: Good comment but suggest to use a destructure instead, we are moving away from argument 2 in `processHtmlAttributes`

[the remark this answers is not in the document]
LuLaValva: Done in dbba32a — destructured, matching the `@image` handling:

```marko
$ const { class: mediaClass, ...mediaInput } = media;
<div ...processHtmlAttributes(mediaInput) class=["card__media", mediaClass]>
```

Verified a consumer `class` now merges (`class="card__media my-media"`) rather than replacing, with other attributes still passing through.

packages/evo-marko/src/tags/evo-card/index.marko
```diff
@@ -44,7 +44,7 @@ export interface Input extends Omit<Marko.HTML.Span, "title"> {
     inputClass,
 ]>
     <div class="card__media">
-        <img ...image>
+        <img ...image class=["card__hero-image", image.class]>
```
HenriqueLimas: do we need to add `@media` for evo-marko also?

packages/evo-marko/src/tags/evo-card/index.marko
```diff
@@ -43,9 +43,15 @@ export interface Input extends Omit<Marko.HTML.Span, "title"> {
...
-    <div class="card__media">
-        <img ...image>
-    </div>
+    <const/{ class: mediaClass, ...mediaInput }=media>
+    <if=media.content>
+        <div ...(mediaInput as Marko.HTML.Div) class=["card__media", mediaClass]/>
+    </if>
+    <else>
+        <div class="card__media">
+            <img ...mediaInput class=["card__hero-image", mediaClass]>
```
HenriqueLimas, line 52: I think this just make it hard for the types, `@media` is now an image if there is no content otherwise it is a completely different component. I think it should be clear to always be one single `div.card__media`. For the image we can introduce a `evo-card-image` mainly for adding the `card__hero-image` or we update skin to look for `.card__media img`

LuLaValva: We use the same in dialog with [banner](https://github.com/eBay/evo-web/blob/main/packages/evo-marko/src/tags/evo-dialog/index.marko#L9-L14), though the pattern is slightly different. I think it's a good idea but I'll try to align more closely to `@banner`

HenriqueLimas: It doesn't seem the same. banner its always a div with a background image. Here its an `img` or a completely different element.

LuLaValva: Hmm, maybe this is a sign that we actually should be doing something more similar to banner then for consistency. I wonder if we could get rid of `"card__hero-image"` if we used background image instead

HenriqueLimas (dismissed):

HenriqueLimas (dismissed):

HenriqueLimas (dismissed):

HenriqueLimas (approved):

PR #887, opened 2026-08-13
docs(evo-react): compact all-icons storybook example into a dense grid

LuLaValva:
evo-react counterpart to #885: replaces the generated all-icons table with a dense grid. Clicking an icon toggles a bubble showing its selectable component tag (`<EvoIconFoo/>`); clicking it again, another icon, or anywhere else closes it. The cells render through a storybook-only `IconExample` helper with CSS-module styles, excluded from the package build (verified absent from `dist/`). The generated stories file joins the existing autogenerated entries in `.prettierignore`.

Dense icon grid with an open bubble showing <EvoIconAiFilled20/>

Storybook-only; no changeset.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #909, opened 2026-08-26
docs(evo-icon): use evo-infotip for icon grid examples in marko and react

LuLaValva:
The dense all-icons storybook grids in evo-marko and evo-react now open a real `evo-infotip` on click instead of `evo-tooltip` (marko) / a hand-rolled bubble (react). The showcased icon becomes the infotip trigger via its custom icon slot, so the grids gain flip/shift at viewport edges and a dismiss button from the component itself. This also restores the evo-react dense grid that was previously pulled from review (#887) while waiting for a suitable overlay component.


packages/evo-react/src/icon/icon-example.tsx
```diff
@@ -0,0 +1,28 @@
...
+  name: string;
+  children: ReactNode;
+}) {
+  return (
+    <div className={styles.cell}>
+      <EvoInfotip
+        className={styles.tip}
+        a11yIconText={name}
+        a11yCloseText="Dismiss"
+        icon={children}
```
HenriqueLimas, line 22: 🤯

PR #910, opened 2026-08-26
fix(skin): use solid tokens for calendar range fill and accent for selected days

LuLaValva:
Calendar day hover and range fills now share one solid background token instead of translucent state layers that stacked, and the selected-day background plus the current-day border and text use the accent color.

### DIFF
--- a/.changeset/calendar-range-solid-fill.md
+++ b/.changeset/calendar-range-solid-fill.md
@@ -2,4 +2,4 @@
 "@ebay/skin": patch
 ---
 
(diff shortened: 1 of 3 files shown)


HenriqueLimas:
It seems that the current date is also blue as per [playbook](https://playbook.ebay.com/design-system/components/calendar/?tab=design), maybe we should do the change in this PR too

Screenshot 2026-08-27 at 4 49 37 PM

HenriqueLimas (approved):

PR #916, opened 2026-08-28
feat(evo-video): custom video component replacing shaka-player

LuLaValva:
Created a new `video` CSS component and `<evo-video>` Marko component using primitives from `videojs` instead of depending on the `<shaka-player>` library.

### Component decisions:
- Closed Captions control is semantically a menu button
- Volume control is **visually hidden**, but secretly always KB accessible and visible to SR

### `<evo-video>` API decisions:
- All inputs are [controllable](https://markojs.com/docs/explanation/controllable-components) (as always in this Tags API migration)
- Whenever possible, `a11y*Text` decides whether an element will be included
  - `cc` requires `a11yCaptionsText`
  - volume has two tiers
    - `a11yMuteText` + `a11yUnmuteText` = just a mute/unmute button
    - `a11yVolumeText` adds a volume slider
  - fullscreen functionality is absent without `a11yFullscreenText`
  - ⭐ The former "compact layout" mode is shown when `a11ySeekText` is absent

### Bonus change
- `<evo-select>` had a few tiny bugs that I surfaced while wrapping it, not worth a separate PR


.marko-run/routes.d.ts
HenriqueLimas, line 1: do we need to have this file in git?

LuLaValva: Typically we recommend this, yes! It allows types to work right away before a `marko-run` build. There are pros and cons though, as it also leads to unnecessary diffs in some PRs

LuLaValva: Yes, typically we include it! Makes it so that types work out of the box, before running a `marko-run` build

HenriqueLimas: Oh probably something that should be fixed in marko run then.

packages/evo-marko/src/tags/evo-video/index.marko
```diff
@@ -0,0 +1,758 @@
+export interface Input extends Marko.HTML.Video {
```
HenriqueLimas, line 1: Should we omit `src` from `HTML.Video` types?

packages/evo-marko/src/tags/evo-video/index.marko
```diff
@@ -0,0 +1,758 @@
...
+  track,
+  nav,
+  playView,
+  autoplay,
+  offscreenPause,
+  spinnerTimeout,
+  errorText = "An error has occurred",
+  report,
+  a11yPlayText = "Play",
+  a11yPauseText = "Pause",
```
HenriqueLimas, line 130: I thought that we decided to not have default hard coded strings, was my understanding incorrect? What are the cases that we should/n't use them?

packages/evo-marko/src/tags/evo-video/index.marko
```diff
@@ -0,0 +1,758 @@
...
+            </button>
+            <span class="video__popover">
+              <evo-menu selected=language || "off">
+                <@item
+                  value="off"
+                  onClick() {
+                    language = null;
+                    languageOpen = false;
+                  }>
+                  Off
```
HenriqueLimas: Should this be hard coded english?

.changeset/add-evo-video.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ebay/skin": patch
```
HenriqueLimas, line 2: skin should be probably a `minor`

```suggestion
"@ebay/skin": minor
```

saiponnada (commented):
Over all LGTM! Few things to take a look, 

- When using AT, over the track, we need to announce timestamp properly . I hear something like `0.2` which is not ideal.
- Weird we lost contrast once cc button is expanded.

Image

- Similarly lets use volume values to be percentage instead of decimals.
- Also wrap the `esc` in abbr tag / or using explicit aria label might be preferred. so AT announces `escape` instead of `e s c`

ArtBlue:
> Volume control is visually hidden, but secretly always KB accessible and visible to SR

Hmm...this means that KB and SR users have a more enhanced UX. Why not just enable it for all users?

LuLaValva:
> > Volume control is visually hidden, but secretly always KB accessible and visible to SR
> 
> Hmm...this means that KB and SR users have a more enhanced UX. Why not just enable it for all users?

KB and SR don't have enhanced UX, volume slider is still reachable by clicking on the volume button. There is not enough room in the menu to always show a volume slider, this approach is standard practice for many video players online

ArtBlue:
> KB and SR don't have enhanced UX, volume slider is still reachable by clicking on the volume button. There is not enough room in the menu to always show a volume slider, this approach is standard practice for many video players online

Ok, a few different terms being used here (some synonymously). We have volume - 

1. button
2. control
3. silder

I was confused by the language. I mistook "control" for the button itself. It looks like you mean it as the "slider." If the button to open the slider is always available, it's all good. Thanks! 😄

packages/skin/src/sass/video/video.scss
```diff
@@ -1,5 +1,281 @@
...
+    position: absolute;
+    width: var(--dimension-800);
+
+    > svg {
+        height: 100%;
+        width: 100%;
+    }
+}
+
+.video--with-nav > a {
```
HenriqueLimas, line 271: Instead of `--with-nav` can we have a class `video__nav`, that helps with avoiding the Children lookup in react
```suggestion
.video__nav {
```

PR #929, opened 2026-09-03
chore(deps): css build toolchain

LuLaValva:
Updates cssnano 9, sass 1.103, postcss-import 17, autoprefixer, and stylelint 17 (no config changes needed). The new minifier merges selectors via :is() and reorders declarations, so the skin dist output is regenerated per repo policy — same styles, new serialization. Percy should confirm no visual change.


HenriqueLimas (changes requested):
The changes of `:is()` is currently breaking percy snapshots in firefox and needs to be investigated

HenriqueLimas (approved):

PR #936, opened 2026-09-08
fix(video): restore compact timestamp styling on span time displays

LuLaValva:
The compact layout’s remaining-time pill was picking up its background, side margins and centering from the `.video-player--compact .shaka-controls-button-panel button` rule. When the time displays were changed from buttons to spans for accessibility, those styles stopped applying and the timestamp rendered as bare text.

The pill styles now also target direct span children of the control panel. Two properties that came from button UA defaults are added explicitly: `box-sizing: border-box` (so the padding stays inside the 48px height / 80px min-width instead of growing the pill to 64×88) and `justify-content: center` (the span is `inline-flex`, so it would otherwise left-align inside its min-width).

Includes the rebuilt `packages/skin/dist/video/video.css`, which is what consumers such as the ebayui-core example actually load.

### DIFF
--- a/.changeset/video-compact-timestamp-styling.md
+++ b/.changeset/video-compact-timestamp-styling.md
@@ -1,0 +1,6 @@
+---
+"@ebay/skin": patch
+---
+
+Restore the compact video layout's timestamp pill styling now that the time
+displays render as spans.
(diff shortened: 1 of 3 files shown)


HenriqueLimas (commented):
we need a changeset

HenriqueLimas (approved):

PR #962, opened 2026-09-16
fix(carousel): make inactive slides inert

### LINKED ISSUE #963: ebay-carousel: VoiceOver reads all slides aria-label at once with links and tabindex: -1
I'm running into a bit of an edge-case with MacOS VoiceOver announcing every alt text for all slides, even the ones not visible in the carousel.

It announces: "list Nike Air Jordan Image 1 of 5 Nike Air Jordan Image 2 of 5 ...etc" instead of just "list Nike Air Jordan Image 1 of 5". Codex suggested setting inert on each out of view li which I tested fixes the issue. We should only add inert to hidden slides when using the carousel with a single visible slide with no peek.

I was able to replicate with some changes to the "Discrete" Storybook example. It requires 2 things:

1. Links
2. tabindex and data-carousel-tabindex -1 on those Links

https://github.com/patrickufer/evo-web/commit/480d14ec55ddea97a407a46c858f94f314565ea8

It's definitely a weird edge-case, but it's necessary for our multi-image carousel for item cards on Search.

Image

Image

https://ebay-eng.slack.com/archives/CK8AX1L49/p1789587695699459

patrickufer:
Closes https://github.com/eBay/evo-web/issues/963

When a single-slide image carousel receives keyboard focus, Chromium can include every image’s alt text in the scroll container’s accessible name—even when inactive slides have aria-hidden="true". VoiceOver consequently announces all image descriptions together.

This change makes inactive items inert when the calculated layout has exactly one item per slide and no adjacent preview:

```
inert=(
    data.itemsPerSlide === 1 &&
    data.peek === 0 &&
    !item.fullyVisible
)
```

Continuous, multi-item, and peek carousels retain their existing behavior.


packages/ebayui-core/src/components/ebay-carousel/index.marko
```diff
@@ -65,6 +65,11 @@ $ var config = data.config;
                         class=item.class
                         style=item.style
                         aria-hidden=!item.fullyVisible && "true"
+                        inert=(
+                            data.itemsPerSlide === 1 &&
```
LuLaValva: Are you sure we only want this functionality when there's 1 item per slide? I suspect this is still a problem with other types of carousels with many items.

Naively, I'd expect `inert=!item.fullyVisible` to be the proper solution

patrickufer: Good call. I think I was being overly cautious, but you're right from this demo with 2 visible slides, we'd want slides 1 and 2 to have `inert`.
image

ianmcburnie:
Yes, nice change, now that `inert` falls within our browser policy we can probably even get rid of the old aria-hidden and tabindex=-1  solutions we had in place. This _should_ work well alongside those in the meantime, but let us check with assistive technology.

packages/ebayui-core/src/components/ebay-carousel/index.marko
```diff
@@ -65,6 +65,7 @@ $ var config = data.config;
                         class=item.class
                         style=item.style
                         aria-hidden=!item.fullyVisible && "true"
+                        inert=!item.fullyVisible
```
saiponnada, line 68: `!item.fullyVisible`  means it also excludes peeked items. what that means is the last item we see with peeks is clickable through mouse but with this change, that is not. That may be intentional, but it’s a behavior change requiring explicit agreement and tests across variants

patrickufer: @saiponnada Good callout. It does seem to align with our directives on peeking carousel items though:

> Peeks must not be focusable or interactive.

https://opensource.ebay.com/evo-web/components/carousel/accessibility#navrail-nav_title

CC @ianmcburnie

patrickufer:
https://issues.chromium.org/issues/563068684

Looks like a Chromium bug. Filed an issue with them. We should still consider `inert` instead of our hacky `aria-hidden` and `tabindex: -1` approach for a future `evo-carousel`, though

